PR #8, opened 2026-03-03
feat: add lessons.md scaffold and seed it to existing projects

ssheld:
## Summary

- Adds a `scaffold/agent-vault/lessons.md` template for tracking mistake patterns and preventive rules across agent sessions
- Updates `AGENTS.md` and `shared-rules.md` scaffolds to reference `lessons.md` in Session Start (read) and Session End (write) rules, plus adds a new "Completion Verification" section and two new Core Principles
- Adds a `seed_if_missing()` helper to `update-project.sh` so existing projects receive the template on their next update — without ever overwriting it afterward
- Updates `README.md` to list `lessons.md` in the project-owned files section

## Motivation

The Session End rule says *"add an entry to `agent-vault/lessons.md`"* without an "if it exists" guard. New projects get `lessons.md` automatically via `cp -R` in `new-project.sh`, but **existing projects** that run `update-project.sh` would not receive the template — agents would need to create the file from scratch with no structure to follow.

The `seed_if_missing()` pattern solves this: it copies the scaffold template only when the destination file does not exist, and silently skips it on subsequent runs. Unlike `sync_managed_file`, it never overwrites — `lessons.md` is project-owned content, not a managed file.

## Changes

| File | Change |
|------|--------|
| `scaffold/agent-vault/lessons.md` | New template with usage rules and empty entries section |
| `scaffold/agent-vault/AGENTS.md` | Add `lessons.md` to Session Start/End rules; add Completion Verification section; add two Core Principles (root causes, plan-before-implement) |
| `scaffold/agent-vault/shared-rules.md` | Mirror the same AGENTS.md updates (these files stay in sync) |
| `scripts/update-project.sh` | Add `seed_if_missing()` function (lines 236-255) and call it for `lessons.md` (line 364) after managed-file syncs |
| `README.md` | List `lessons.md` in the project-owned files section |

## How `seed_if_missing()` works

```bash
seed_if_missing "$vault_scaffold_dir/lessons.md" "$project_dir/lessons.md"
```

1. If `$dest` already exists → silently return (no-op)
2. If `--dry-run` → print `Seed: agent-vault/lessons.md (new template)`
3. Otherwise → copy the scaffold template and print `Seeded: agent-vault/lessons.md (new template)`
4. Increments the `created` counter in either active case

Key design decisions:
- Placed **after** `sync_managed_file` calls and **before** `ensure_obsidian_gitignore` in the execution order
- Uses the same `repo_relative_path()` helper for consistent output formatting
- Respects `--dry-run` mode
- Never overwrites existing content — project-owned, not managed

## Test plan

- [ ] Run `./scripts/update-project.sh <repo-path> --dry-run` on a project that lacks `lessons.md` — should print `Seed: agent-vault/lessons.md (new template)`
- [ ] Run again without `--dry-run` — should print `Seeded: agent-vault/lessons.md (new template)` and the file should exist with the template content
- [ ] Run a third time — `lessons.md` exists now, so `seed_if_missing` silently returns with no output for that file
- [ ] Manually add content to `lessons.md` in the target project, run update again — content is preserved (never overwritten)
- [ ] Run `bash -n scripts/update-project.sh` — syntax check passes
- [ ] Verify `new-project.sh` still works (no changes needed — `cp -R` already copies the entire scaffold)

### DIFF
--- a/scripts/update-project.sh
+++ b/scripts/update-project.sh
@@ -110,7 +110,8 @@ for required in \
   "$vault_scaffold_dir/CLAUDE.md" \
   "$vault_scaffold_dir/GEMINI.md" \
   "$vault_scaffold_dir/shared-rules.md" \
-  "$vault_scaffold_dir/review-policy.md"
+  "$vault_scaffold_dir/review-policy.md" \
+  "$vault_scaffold_dir/lessons.md"
 do
   if [[ ! -f "$required" ]]; then
     echo "Error: missing scaffold file: $required"
@@ -239,6 +240,12 @@ seed_if_missing() {
   local rel
 
   rel="$(repo_relative_path "$dest")"
+
+  if [[ -L "$dest" ]]; then
+    echo "Skip: $rel (symlink — not seeding)"
+    skipped=$((skipped + 1))
+    return
+  fi
 
   if [[ -e "$dest" ]]; then
     return


ssheld:

Overall this is a good improvement (especially seeding `lessons.md` as project-owned content), but I found one blocking safety issue and one reliability gap.

**Critical**
- `scripts/update-project.sh` (`seed_if_missing`) can write through dangling symlinks.
- Current logic checks `[[ -e "$dest" ]]`, but a dangling symlink makes `-e` false while `cp "$src" "$dest"` still follows the symlink target.
- This allows writes outside the repo path if `agent-vault/lessons.md` is a dangling symlink.
- Please add a symlink guard (`[[ -L "$dest" ]]`) and fail with a clear message, or add equivalent preflight protection.

**Recommended**
- Add `scaffold/agent-vault/lessons.md` to the required scaffold preflight list.
- Without this, a missing scaffold file is detected late (at seed time) after other files may already have been updated, causing partial update state.

**Merge recommendation:** Request Changes  
**Top risks:**
- Arbitrary file write path via dangling symlink in seed flow
- Partial updates if scaffold `lessons.md` is missing

**Suggested additional tests:**
- `update-project.sh --dry-run` and real run when `agent-vault/lessons.md` is a dangling symlink (expect safe failure)
- Missing `scaffold/agent-vault/lessons.md` should fail before any managed file updates

PR #28, opened 2026-03-08
Fix managed .gitignore comment deduplication

### LINKED ISSUE #25: Deduplicate managed .gitignore comments when ignore patterns already exist
## Summary
The managed `.gitignore` helper currently treats comment lines as independent required entries. In repos that already contain the Obsidian ignore patterns but not the newer scaffold comments, a sync can append orphaned comments after the existing block.

Example result:
```gitignore
.obsidian/cache/
.obsidian/backup/
.obsidian/plugins/*/data.json
# Obsidian -- machine-specific & volatile files (ignore these)
# Plugin data (can contain API keys or large caches)
# Agent Vault -- local sync and migration backups (ignore these)
/agent-vault/context/updates/
```

Only the Agent Vault comment + pattern are actually new there. The two Obsidian comments become dangling duplicates.

## Why this matters
- Creates noisy `.gitignore` diffs in synced repos.
- Makes the managed ignore helper look sloppy even though the actual patterns are correct.
- The issue can recur anywhere the Obsidian patterns predate the newer scaffold comment block.

## Suggested direction
- Make the `.gitignore` helper comment-aware instead of treating comments as standalone required lines.
- At minimum, avoid appending the Obsidian comment lines when their corresponding ignore patterns are already present.
- Ideally, group comment + pattern insertion so the resulting block stays coherent.

## Affected paths
- `scripts/new-project.sh`
- `scripts/update-project.sh`

## Context
- Surfaced during review of `north-ai-website` PR #10 after syncing the latest scaffold changes.

ssheld:
## Summary
Fix the managed `.gitignore` helper so it inserts comment + pattern blocks coherently instead of treating every managed line as an independent required entry.

Closes #25.

## Files Changed
- `scripts/new-project.sh`
  - Replace the flat managed `.gitignore` line list with block-aware insertion logic.
  - Avoid appending dangling comment lines when the corresponding ignore patterns already exist.
- `scripts/update-project.sh`
  - Apply the same block-aware `.gitignore` behavior to updates for existing repos.
- `scripts/test-gitignore-management.sh`
  - Add regression coverage for:
    - existing legacy Obsidian patterns without the newer managed comments
    - partially present Obsidian blocks that still need grouped insertion
    - both `new-project.sh` and `update-project.sh`

## Validation
- `bash scripts/test-gitignore-management.sh` — passed
- `bash -n scripts/new-project.sh scripts/update-project.sh scripts/test-gitignore-management.sh` — passed
- `git diff --check` — passed

## Risks / Rollback
- Low risk: behavior change is narrowly scoped to managed `.gitignore` insertion.
- The main tradeoff is that comments are now added only when at least one ignore pattern from that block is missing, which is the intended fix for noisy orphaned comments.
- Rollback is straightforward: revert this commit if a different `.gitignore` management strategy is preferred.

## Docs
Documentation is consistent - design.md and README.md were checked and need no changes.

### DIFF
--- a/scripts/new-project.sh
+++ b/scripts/new-project.sh
@@ -123,6 +123,9 @@ append_migrated_root_content() {
   } >> "$destination_path"
 }
 
+# Contract: each block starts with a human-readable comment line followed by
+# one or more ignore patterns. collect_missing_managed_gitignore_lines relies
+# on that shape when deciding whether to add the block comment.
 MANAGED_GITIGNORE_BLOCKS=(
   $'# Obsidian -- machine-specific & volatile files (ignore these)\n.obsidian/workspace.json\n.obsidian/app.json\n.obsidian/appearance.json\n.obsidian/workspace-mobile.json\n.obsidian/cache/\n.obsidian/backup/'
   $'# Plugin data (can contain API keys or large caches)\n.obsidian/plugins/*/data.json'
--- a/scripts/update-project.sh
+++ b/scripts/update-project.sh
@@ -31,6 +31,9 @@ expand_path() {
   esac
 }
 
+# Contract: each block starts with a human-readable comment line followed by
+# one or more ignore patterns. collect_missing_managed_gitignore_lines relies
+# on that shape when deciding whether to add the block comment.
 MANAGED_GITIGNORE_BLOCKS=(
   $'# Obsidian -- machine-specific & volatile files (ignore these)\n.obsidian/workspace.json\n.obsidian/app.json\n.obsidian/appearance.json\n.obsidian/workspace-mobile.json\n.obsidian/cache/\n.obsidian/backup/'
   $'# Plugin data (can contain API keys or large caches)\n.obsidian/plugins/*/data.json'


ssheld:
## Review

Nice fix — the block-aware approach is a clean solution to the orphaned comment problem, and the regression tests are thorough. A few observations:

### Code duplication between scripts

`collect_missing_managed_gitignore_lines()` and `MANAGED_GITIGNORE_BLOCKS` are duplicated nearly identically between `new-project.sh` and `update-project.sh`. If these drift apart it'll be a maintenance headache. Consider extracting the shared data and function into a sourced helper (e.g. `scripts/lib/gitignore-helpers.sh`) that both scripts `source`.

Similarly, `gitignore_has_line()` now has two slightly different implementations — the one in `new-project.sh` uses inline `awk` directly, while `update-project.sh` delegates to `has_exact_line`. Worth unifying.

### Block format is a fragile implicit contract

The logic assumes the first line of every block is always a comment (it's captured as `comment_line` and skipped from pattern checking). If someone later adds a block without a leading comment, the first actual pattern would be silently dropped. A brief comment above `MANAGED_GITIGNORE_BLOCKS` documenting this contract would help, or a guard like:

```bash
if [[ "$comment_line" != \#* ]]; then
  echo "Error: block must start with a comment line" >&2
  return 1
fi
```

### Block definition readability

The `$'...\n...\n...'` single-line block definitions are dense and hard to scan. If you're not inclined to extract them to a data file, even just splitting them across lines with `\` continuations would help readability.

### Minor: `printf` on empty array

When `missing_lines` is empty, `printf '%s\n' "${missing_lines[@]}"` outputs a single empty line. The consumer handles this with the `[[ -n "$line" ]]` guard so it works fine — just worth being aware of.

### Tests

Good coverage — testing both scripts against both legacy (all patterns present) and partial scenarios is exactly right. The `assert_line_count` helper for dedup verification is a nice touch. The cleanup trap is properly set.

**Overall: LGTM with the duplication as the main thing worth addressing (now or in a follow-up).**

PR #52, opened 2026-03-16
fix: Make hook activation failures visible in scaffold scripts

ssheld:
## Summary

- `configure_tracked_hooks_path()` silently returned 0 even when it failed to activate the hook (e.g., custom `core.hooksPath` conflict or not-a-repo). Callers had no way to detect failure.
- `update-project.sh` printed the summary *before* calling hook activation, so the output gave no visibility into whether it succeeded.
- This caused CritterCam's PR #83 scaffold sync to silently skip hook activation, with the daily note and design-log incorrectly claiming it was enabled.

### `scripts/lib/tracked-hooks.sh`
- Return 0 only on confirmed activation (set, already set, or dry-run).
- Return 1 on skip (not a repo, custom path conflict) or failure.
- Added read-back verification after `git config --local` to catch silent write failures.

### `scripts/update-project.sh`
- Moved `configure_tracked_hooks_path` call before the summary block.
- Captures return code and includes hook status in the summary output.
- On failure, prints a warning with the manual activation command.

### `scripts/new-project.sh`
- Captures return code from `configure_tracked_hooks_path`.
- Prints warning with manual command on failure.

### `scripts/test-session-metadata-hook.sh`
- Added test: custom `core.hooksPath` → `configure_tracked_hooks_path` returns non-zero.
- Added test: fresh activation → returns zero and sets the expected path.

## Validation
- `bash scripts/test-session-metadata-hook.sh` → all regression checks passed.

## Risks
- Callers that previously ignored the return code will now see a non-zero exit. Since both callers (`update-project.sh`, `new-project.sh`) run under `set -e`, the new explicit return-code capture prevents unexpected script termination.

Documentation is consistent — no design.md or README.md changes needed (shell scripts only).

### DIFF
--- a/scripts/lib/tracked-hooks.sh
+++ b/scripts/lib/tracked-hooks.sh
@@ -7,7 +7,7 @@ configure_tracked_hooks_path() {
   local current_hooks_path=""
 
   if ! git -C "$repo_root" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
-    return
+    return 1
   fi
 
   current_hooks_path="$(git -C "$repo_root" config --get core.hooksPath 2>/dev/null || true)"
@@ -17,17 +17,22 @@ configure_tracked_hooks_path() {
       echo "Dry run: would enable tracked metadata hook via core.hooksPath=$desired_hooks_path"
     else
       git -C "$repo_root" config --local core.hooksPath "$desired_hooks_path"
+      if [[ "$(git -C "$repo_root" config --get core.hooksPath 2>/dev/null)" != "$desired_hooks_path" ]]; then
+        echo "Warning: core.hooksPath was set but read-back does not match expected value '$desired_hooks_path'." >&2
+        return 1
+      fi
       echo "Enabled tracked metadata hook via core.hooksPath=$desired_hooks_path"
     fi
-    return
+    return 0
   fi
 
   if [[ "$current_hooks_path" == "$desired_hooks_path" ]]; then
     echo "Tracked metadata hook already enabled via core.hooksPath=$desired_hooks_path"
-    return
+    return 0
   fi
 
   echo "Notice: effective core.hooksPath already set to '$current_hooks_path'; left unchanged." >&2
   echo "To use the tracked agent-vault hook in this clone, run:" >&2
   echo "  git -C \"$repo_root\" config core.hooksPath $desired_hooks_path" >&2
+  return 1
 }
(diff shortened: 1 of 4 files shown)


ssheld (commented):

1. `scripts/update-project.sh:523`

   **Recommended** · High confidence — Dry-run summary reports hook activation even when nothing was changed.
   With `--dry-run` and no existing `core.hooksPath`, `configure_tracked_hooks_path` now returns success for the "would enable" branch, so the new summary prints `- hook: enabled` even though the command later says `Dry run complete. No files were written.` I reproduced this on the PR branch in a fresh repo. The summary should distinguish "already enabled" from "would enable" so the status line stays truthful.

## Strengths
- Hook activation failures are no longer silently swallowed by the callers.
- The added regression cases cover both the custom-hook conflict and the successful activation path.

## Merge Recommendation
Approve with Changes

## Top Risks
- `update-project.sh --dry-run` can still mislead users into thinking the metadata hook is active when the config was never written.

## Suggested Additional Tests
- Add a dry-run regression for `update-project.sh` that asserts the summary says `would enable` or otherwise marks the hook as not yet active when `core.hooksPath` is unset.

ssheld (commented):

1. `scripts/update-project.sh:523`

   **Recommended** · High confidence — The new dry-run status still misreports the already-enabled path.
   The follow-up commit fixes the unset-hook case by printing `would enable` during dry-run, but `configure_tracked_hooks_path` also returns success when the hook is already configured. Because the summary now keys only on `hook_rc == 0 && dry_run == true`, `update-project.sh --dry-run` prints `- hook: would enable` even in repos where `core.hooksPath` is already `agent-vault/_assets/hooks`. I reproduced this on the current head in a freshly bootstrapped repo.

## Strengths
- The original unset-hook dry-run case is now reported more accurately.
- The new helper regression test keeps the non-writing dry-run behavior explicit.

## Merge Recommendation
Approve with Changes

## Top Risks
- Dry-run output can still mislead users about whether the tracked hook is already active versus only planned.

## Suggested Additional Tests
- Add an `update-project.sh --dry-run` regression for a repo where `core.hooksPath` is already `agent-vault/_assets/hooks`, and assert the summary reports `enabled` or `already enabled` rather than `would enable`.

ssheld (commented):

## Strengths
- `48dd740` now distinguishes `enabled` from `would enable` by reading the repo's effective `core.hooksPath`, which addresses the prior dry-run reporting bug.
- I re-ran `bash scripts/test-session-metadata-hook.sh`, and it passed on the current head.
- I also verified `update-project.sh --dry-run` in both relevant cases on the current head: already-enabled repos report `- hook: enabled`, and unset-hook repos report `- hook: would enable`.

## Merge Recommendation
Approve

## Top Risks
- No new high-confidence risks identified in the follow-up patch.

## Suggested Additional Tests
- Optional: add an end-to-end `update-project.sh --dry-run` assertion for the already-enabled path so the summary text is covered directly, not only through helper behavior.

PR #80, opened 2026-03-18
Enforce context-log freshness in tracked metadata hook

### LINKED ISSUE #79: Enforce context-log ordering and freshness in tracked metadata hook
> 🤖 **Post by Codex** · via Codex CLI

## Summary
Repeated downstream PR feedback shows that `agent-vault/context-log.md` can drift even when contributors follow the baseline metadata gate. The scaffold already documents `Newest entry at top.` in the context-log template, but the tracked `pre-commit` hook only checks that `context-log.md`, a daily note, and a design-log note are staged. It does not validate whether the staged context log is internally consistent.

That leaves a gap where a newer entry can be appended below older entries, or where `last_updated` / Current Snapshot remain stale after a new entry is added. Reviewers then have to catch this manually.

## Downstream Evidence
- `ssheld/critter-cam#112` surfaced a concrete example: a March 18 entry was appended below older March 9 notes, which left the top-of-file snapshot stale even though the right files were staged.
- This does not appear to be a missing policy rule. The rule already exists in `scaffold/agent-vault/context-log.md` and `scaffold/agent-vault/Templates/Context Log.md`.
- The underlying gap is enforcement in the tracked hook and regression coverage.

## Proposed Fix
Extend `scaffold/agent-vault/_assets/hooks/pre-commit` so that when `agent-vault/context-log.md` is staged for a substantive commit, it also validates:
- entry headings use a machine-checkable format such as `YYYY-MM-DD HH:MM local - <agent> - <topic>`
- entries remain newest-first
- frontmatter `last_updated` matches the top entry date
- Current Snapshot `Last updated` matches the top entry date

Add regression coverage so generated repos inherit this behavior and future scaffold changes cannot regress it silently.

## Why This Matters
`context-log.md` is the canonical cross-session memory file. If the newest state is not actually at the top, later agents can continue from stale goals, milestone guidance, or handoff references even though the repo technically contains the newer note farther down.

## Acceptance Criteria
- The tracked hook rejects a staged `context-log.md` when a newer entry appears below an older one.
- The tracked hook rejects a staged `context-log.md` when `last_updated` or Current Snapshot `Last updated` are stale relative to the top entry.
- The context-log template documents the enforced heading format.
- Regression tests cover both ordering and freshness failures.

ssheld:
## Summary
- validate staged `agent-vault/context-log.md` ordering and freshness in the tracked `pre-commit` hook
- document the enforced context-log entry heading format in the scaffold and hook docs
- add regression coverage for newest-first ordering and stale `last_updated` failures

Closes #79.

## Testing
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ When a managed file changes, the script backs up the previous version under:
 Generated projects auto-enable the tracked metadata gate in the clone where `new-project.sh` or `update-project.sh` runs, unless `core.hooksPath` is already set to something else. Additional clones can enable it with:
 - `git -C <repo-path> config core.hooksPath agent-vault/_assets/hooks`
 
+That tracked hook enforces the baseline session artifacts and validates staged `agent-vault/context-log.md` ordering/freshness so newer entries do not get appended below stale headers.
+
 Both scripts also ensure root `.gitignore` includes managed local-only ignore entries (added only when missing):
 - `.obsidian/workspace.json`
 - `.obsidian/app.json`
--- a/scaffold/agent-vault/Templates/Context Log.md
+++ b/scaffold/agent-vault/Templates/Context Log.md
@@ -8,6 +8,7 @@ last_updated:
 
 ## Usage Rules
 - Newest entry at top.
+- Entry headings must start with `YYYY-MM-DD HH:MM local - <agent> - <topic>`.
 - Keep entries short and concrete.
 - Reference files and PRs instead of pasting long diffs.
 - Pair each major update with a design-log entry.
--- a/scaffold/agent-vault/_assets/hooks/README.md
+++ b/scaffold/agent-vault/_assets/hooks/README.md
@@ -17,6 +17,10 @@ git config core.hooksPath agent-vault/_assets/hooks
     - `agent-vault/context-log.md`
     - one note under `agent-vault/daily/`
     - one note under `agent-vault/design-log/`
+  - Validates staged `agent-vault/context-log.md` content:
+    - entry headings must use `YYYY-MM-DD HH:MM local - <agent> - <topic>`
+    - entries must remain newest-first
+    - frontmatter and Current Snapshot `Last updated` must match the top entry date
   - This is a baseline gate only. Conditional artifacts such as `open-questions.md`, decision records, handoff notes, and `lessons.md` still depend on the actual session outcome.
 
 ## Intentional Bypass
--- a/scaffold/agent-vault/_assets/hooks/pre-commit
+++ b/scaffold/agent-vault/_assets/hooks/pre-commit
@@ -46,6 +46,143 @@ is_runtime_note_file() {
   esac
 }
 
+extract_context_log_frontmatter_value() {
+  local file_path="$1"
+  local field_name="$2"
+
+  awk -v field_name="$field_name" '
+    BEGIN {
+      frontmatter_markers = 0
+    }
+    /^---$/ {
+      frontmatter_markers += 1
+      next
+    }
+    frontmatter_markers == 1 && index($0, field_name ":") == 1 {
+      value = substr($0, length(field_name) + 2)
+      sub(/^[[:space:]]+/, "", value)
+      sub(/[[:space:]]+$/, "", value)
+      print value
+      exit
+    }
+    frontmatter_markers >= 2 {
+      exit
+    }
+  ' "$file_path"
+}
+
+extract_context_log_snapshot_last_updated() {
+  local file_path="$1"
+
+  awk '
+    /^## Current Snapshot$/ {
+      in_snapshot = 1
+      next
+    }
+    in_snapshot && /^## / {
+      exit
+    }
+    in_snapshot && /^- Last updated:/ {
+      value = $0
+      sub(/^- Last updated:[[:space:]]*/, "", value)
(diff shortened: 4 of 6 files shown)


ssheld (commented):
> 🤖 **Review by Claude Opus 4.6** · via Claude Code

1. `scaffold/agent-vault/_assets/hooks/pre-commit:120`

   **Recommended** · High confidence — Temp file not cleaned up on unexpected exit.

   `validate_staged_context_log` creates a temp file via `mktemp` but has no `trap` to clean it up if the hook is killed or hits an unexpected error between creation and the `rm -f` at line 171. A one-line trap after the `mktemp` would prevent orphaned files in `/tmp`:

   ```bash
   trap 'rm -f "$tmp_file"' RETURN
   ```

2. `scripts/test-session-metadata-hook.sh`

   **Recommended** · Medium confidence — No test for invalid entry heading format.

   The `collect_context_log_entry_timestamps` function emits `__INVALID__` lines for malformed headings, and `validate_staged_context_log` reports them with a specific error message. Neither the ordering nor freshness test exercises this path. A short test case with a heading like `### bad heading format` would close the gap.

3. `scripts/test-session-metadata-hook.sh`

   **Optional** · Medium confidence — No test for empty entries section.

   The `has_entries` guard (pre-commit line 158) errors when `## Entries` contains no entry headings. This path isn't covered by the test suite. Lower priority since the scaffold always seeds an entry, but it would harden the validation against future template changes.

## Strengths
- Clean decomposition into small, well-named helper functions (`extract_context_log_frontmatter_value`, `collect_context_log_entry_timestamps`, etc.) — easy to follow and maintain.
- Validates against the staged blob via `git show ":$file_path"` rather than the working tree, which is the correct approach for a pre-commit hook.
- Error messages are specific and actionable — they tell the user exactly what's wrong and what the expected value should be.

## Merge Recommendation
Approve with Changes

## Top Risks
- Orphaned temp files under sustained hook failures (finding 1).
- The `__INVALID__` heading code path is untested and could silently regress.

## Suggested Additional Tests
- A test with a malformed entry heading (e.g., `### free-form text without timestamp`) to exercise the `__INVALID__` error path.
- A test with an empty `## Entries` section to cover the `has_entries` guard.

ssheld:
> 🤖 **Feedback response by Codex** · via Codex CLI

| 1 | Codex P1 — metadata-only `context-log.md` commits bypassed the new ordering/freshness validation because the hook exited before `validate_staged_context_log` | Resolved | `06534fa`; `scaffold/agent-vault/_assets/hooks/pre-commit`, `scripts/test-session-metadata-hook.sh`; moved context-log validation ahead of the metadata-only early exit and added a metadata-only stale-header regression case. |
| 2 | Claude — temp file created by `mktemp` was not protected against unexpected exit | Resolved | `06534fa`; `scaffold/agent-vault/_assets/hooks/pre-commit`; added a `RETURN` trap around the staged context-log temp file lifecycle so unexpected function exits do not leave orphaned temp files behind. |
| 3 | Claude — invalid heading format path was untested | Resolved | `06534fa`; `scripts/test-session-metadata-hook.sh`; added a malformed heading regression that exercises the `__INVALID__` path and asserts the expected error message. |
| 4 | Claude — empty `## Entries` path was untested | Resolved | `06534fa`; `scripts/test-session-metadata-hook.sh`; added an empty-entries regression that verifies the hook rejects a staged context log with no entry headings. |

## Remaining Items
- None.

## Validation
- `git diff --check`

ssheld (commented):
> 🤖 **Review by Claude Opus 4.6** · via Claude Code

All prior feedback has been addressed in `06534fa`:

| 1 | Temp file not cleaned up on unexpected exit | Resolved | `06534fa`; `scaffold/agent-vault/_assets/hooks/pre-commit`; added `trap 'rm -f "$tmp_file"; trap - RETURN' RETURN` after `mktemp`, with an explicit `trap - RETURN` and `rm -f` before normal return. Clean lifecycle. |
| 2 | No test for invalid entry heading format | Resolved | `06534fa`; `scripts/test-session-metadata-hook.sh`; new `invalid_heading_repo` test case with `replace_first_context_log_heading` helper exercises the `__INVALID__` path and asserts the expected error message. |
| 3 | No test for empty entries section | Resolved | `06534fa`; `scripts/test-session-metadata-hook.sh`; new `empty_entries_repo` test case with `clear_context_log_entries` helper verifies the `has_entries` guard rejects a context log with no entry headings. |
| Codex P1 | Metadata-only context-log commits bypassed validation | Resolved | `06534fa`; `scaffold/agent-vault/_assets/hooks/pre-commit`; moved `non_deleted_staged_files` detection and context-log validation ahead of the metadata-only early exit so validation runs unconditionally. New `metadata_only_invalid_repo` test case confirms stale headers are caught even without non-metadata files staged. |

The restructuring is clean — `non_deleted_staged_files` and `has_context_log` are now computed once early and reused by both the validation block and the later metadata gate loop, with no duplication.

## Strengths
- All four findings addressed in a single, focused commit.
- The structural fix for the Codex P1 (moving validation before the early exit) is the right approach — no special-casing needed.
- Four new test cases bring solid coverage across the invalid heading, empty entries, metadata-only stale header, and ordering scenarios.

## Merge Recommendation
Approve

## Top Risks
- None identified in the current diff.

## Suggested Additional Tests
- None.

PR #85, opened 2026-03-25
Auto-migrate legacy context logs during scaffold sync

### LINKED ISSUE #84: Auto-migrate legacy context logs when syncing the stricter metadata hook
> 🤖 **Post by Codex** · via Codex CLI

## Summary

`update-project.sh` now syncs a stricter `agent-vault/_assets/hooks/pre-commit` that validates staged `agent-vault/context-log.md` structure, but it does not migrate older generated repos whose `context-log.md` still uses the pre-validator layout.

That means a generated repo can land in this state:
- sync succeeds
- managed hook is updated
- next substantive metadata-gated commit fails until someone manually reshapes `agent-vault/context-log.md`

## Problem

The new hook expects a canonical top-level block shaped like:
- `## Current Snapshot`
- `## Entries`
- top entry date matching frontmatter `last_updated`
- `Current Snapshot` `Last updated` matching that same top entry date

But `scripts/update-project.sh` currently only syncs the hook and managed docs/templates. It does **not** detect or migrate legacy `agent-vault/context-log.md` files that predate that validator contract.

So the upgrade path for older generated repos is incomplete: the scaffold can install a stricter gate without also making the repo compatible with that gate.

## Concrete Repro

1. Start from an older generated repo whose `agent-vault/context-log.md` predates the new validator layout.
2. Run `scripts/update-project.sh <repo> --sync-templates`.
3. Make a substantive change and stage the required metadata artifacts.
4. Stage `agent-vault/context-log.md`.
5. `agent-vault/_assets/hooks/pre-commit` can fail because the synced validator now expects the newer top-level `Current Snapshot` / `Entries` layout.

I hit exactly this while syncing `ssheld/critter-cam` in PR `#194`, where I had to add a manual compatibility block to keep the migrated hook usable.

## Proposed Direction

Preferred fix:
- teach `scripts/update-project.sh` to detect legacy `agent-vault/context-log.md` layout and auto-insert the minimal compatibility block needed by the validator, while preserving historical content below it

Minimum acceptable fallback:
- if auto-migration is too risky, emit a loud post-sync warning when the stricter hook is installed into a repo with legacy `context-log.md` structure, including the exact remediation steps

Follow-up hardening:
- add a regression test that seeds an older generated repo, runs `update-project.sh`, and proves the next metadata-gated substantive commit still works

## Acceptance Criteria

- [ ] `scripts/update-project.sh` does not leave older generated repos in a “sync succeeded, next metadata-gated commit fails” state
- [ ] Legacy generated repos either receive an automatic `context-log.md` compatibility migration or a clear sync-time warning with exact remediation
- [ ] Regression coverage exists for syncing the stricter hook into a legacy generated repo

## Related Files

- `scripts/update-project.sh`
- `scaffold/agent-vault/_assets/hooks/pre-commit`
- `scaffold/agent-vault/context-log.md`
- `scaffold/agent-vault/Templates/Context Log.md`

## Related

- https://github.com/ssheld/critter-cam/pull/194

ssheld:
## Summary
- auto-migrate recognized legacy `agent-vault/context-log.md` layouts before syncing the stricter tracked hook
- warn and no-op on missing or unrecognized layouts instead of guessing
- add regression coverage and a real legacy fixture for the migration path

## Verification
- `bash -n scripts/update-project.sh scripts/test-session-metadata-hook.sh scaffold/agent-vault/_assets/hooks/pre-commit`
- `./scripts/test-session-metadata-hook.sh`
- `./scripts/check-policy-mirrors.sh`
- `git diff --check`

Fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ Generated projects auto-enable the tracked metadata gate in the clone where `new
 - `git -C <repo-path> config core.hooksPath agent-vault/_assets/hooks`
 
 That tracked hook enforces the baseline session artifacts and validates staged `agent-vault/context-log.md` ordering/freshness so newer entries do not get appended below stale headers.
+When syncing older generated repos, `update-project.sh` now auto-migrates recognized legacy `agent-vault/context-log.md` layouts into the validator-compatible top-level `## Current Snapshot` / `## Entries` shape before syncing the stricter hook. If the layout is not recognized, the script leaves the file unchanged and prints a manual-remediation warning instead of guessing.
 
 Both scripts also ensure root `.gitignore` includes managed local-only ignore entries (added only when missing):
 - `.obsidian/workspace.json`
--- a/scaffold/agent-vault/_assets/hooks/pre-commit
+++ b/scaffold/agent-vault/_assets/hooks/pre-commit
@@ -242,8 +242,6 @@ has_design_log="false"
 
 for file_path in "${non_deleted_staged_files[@]}"; do
   case "$file_path" in
-    agent-vault/context-log.md)
-      ;;
     agent-vault/daily/*.md)
       if [[ "$file_path" != "agent-vault/daily/README.md" ]]; then
         has_daily_note="true"
--- a/scripts/test-fixtures/context-log/legacy-known.md
+++ b/scripts/test-fixtures/context-log/legacy-known.md
@@ -1,0 +1,51 @@
+---
+type: context-log
+project: fixture-project
+last_updated: 2026-03-25
+---
+
+# Context Log
+
+### 2026-03-25 12:40 local — tightened PR #193 after review feedback
+#### Goal
+Preserve a real legacy entry shape with the older em-dash separator.
+
+#### What Changed
+- Stored current-session entries before the late `Current Snapshot` / `Entries`
+  block.
+
+#### Verification
+- Read-back check only.
+
+#### References
+- `agent-vault/context-log.md`
+
+## Current Snapshot
+- Project: fixture-project
+- Primary goal: Preserve older generated context-log layout for migration
+  testing.
+- Current status: This legacy fixture keeps the snapshot/index block near the
+  bottom of the file.
+- Active branch: `main`
+- Last updated: 2026-03-25
+
+## Entries
+
+### 2026-03-21 19:43 local - codex - older indexed entry
+#### Goal
+Preserve a historical indexed entry below the late `## Entries` heading.
+
+#### State
+- This entry exists to prove the migration preserves the older indexed section.
+
+#### Decisions
+- None.
+
+#### Open Questions
+- None.
+
+#### Next Prompt
+"Finish migrating the context log."
+
+#### References
+- `agent-vault/context-log.md`
--- a/scripts/test-session-metadata-hook.sh
+++ b/scripts/test-session-metadata-hook.sh
@@ -34,6 +34,18 @@ assert_file_exists() {
   fi
 }
 
+assert_files_equal() {
+  local left="$1"
+  local right="$2"
+
+  if ! cmp -s "$left" "$right"; then
+    echo "Expected files to match:" >&2
+    echo "  $left" >&2
+    echo "  $right" >&2
+    exit 1
+  fi
+}
+
 assert_executable() {
   local file_path="$1"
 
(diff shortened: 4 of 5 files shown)


ssheld (commented):
> 🤖 **Review by Claude Opus 4.6** · via Claude Code
Formal review state unavailable: self-review limitation

I traced through the classifier, migration logic, and test scenarios in detail. This is a solid implementation of the agreed plan from #84.

**Classifier verification (`classify_context_log_layout`):**
- `missing`: file doesn't exist → no-op. ✓
- `current`: `## Current Snapshot` before `## Entries`, and either no entry headings exist, or they come after `## Entries`, or historical wrappers are present (already migrated). ✓
- `legacy-known`: entry headings (including at least one em-dash heading) appear before `## Current Snapshot`, which appears before `## Entries`, no historical wrappers. ✓
- `unknown`: everything else → warning + no-op. ✓

The em-dash requirement for `legacy-known` is the right conservative choice — it limits auto-migration to the specific known scaffold-emitted pattern. Repos with only regular-hyphen entries before the snapshot get the `unknown` warning, which is safer than guessing.

**Date consistency in migration (`migrate_legacy_context_log_if_needed`):**
- Frontmatter `last_updated`: `$today` ✓
- Current Snapshot `Last updated`: `$today` ✓
- Migration entry timestamp: `$now_local` (whose date portion = `$today`) ✓

All three derive from `date` calls at script startup (lines 148-149). The validator will see matching dates. Verified that the end-to-end test (`run_hook_expect_success` after staging the migrated file) would catch any mismatch.

**Migration placement (line 851):**
`migrate_legacy_context_log_if_needed` runs after managed docs are synced but **before** `_assets/hooks/pre-commit` is synced. This matches the plan discussion — the context-log is validator-compatible before the stricter hook arrives on disk. ✓

**Scaffold hook fix (line 242-243):**
Removes the dead empty `agent-vault/context-log.md)` case branch from the scaffold's `pre-commit`. This was already validated in the generated CritterCam repo (PR ssheld/critter-cam#194); now the scaffold source matches. ✓

**Test coverage (4 new scenarios):**
1. `missing` — file deleted before sync → no migration output, no warning. ✓
2. `legacy-known` dry-run — file unchanged, output shows what would happen. ✓
3. `legacy-known` real migration — file restructured, all historical sections present, then a substantive staged change passes the hook end-to-end. ✓ (This is the critical test.)
4. `unknown` — unrecognized layout left unchanged, skip + warning messages verified. ✓

**Test fixture (`scripts/test-fixtures/context-log/legacy-known.md`):**
Modeled on the real pre-migration CritterCam layout: em-dash entry heading before `## Current Snapshot`, snapshot and entries at the bottom, regular-hyphen entry under `## Entries`. The fixture's hardcoded `last_updated: 2026-03-25` doesn't affect the test because the migration overwrites frontmatter with the runtime `$today`.

**Minor observations (sub-threshold):**

- If a legacy context-log lacks the `# Context Log` heading entirely, `prefix_start_line` defaults to 1 and the frontmatter would be captured into `## Legacy Unindexed Entries` alongside the entry content (cosmetic duplication, since migration writes fresh frontmatter). Unlikely for scaffold-generated files since the heading has always been present.
- `today` and `now_local` are computed once at script startup. A sync that crosses midnight would produce a stale date in the migration entry. Negligible for a script that completes in seconds.

## Strengths
- Directly implements the plan agreed in #84, including all six feedback points from the discussion (migration-before-hook-sync, four-bucket classifier, concrete legacy-known heuristics, single-source date, warning + no-op for unknown, real fixture data).
- The `replace_file_from_tmp` helper reuses existing safety infrastructure (`validate_write_path`, `assert_not_symlink`, backup semantics, dry-run support) rather than reimplementing it.
- The end-to-end test (migrate → stage substantive change → hook passes) is the most valuable test case and exercises the full integration path.

## Merge Recommendation
Approve

## Top Risks
- Legacy repos that used only regular-hyphen entries before the snapshot (no em-dashes) will get the `unknown` warning instead of auto-migration. This is the correct conservative behavior, but could generate support questions from repos that expect auto-migration. The warning message includes clear remediation steps.
- The migration is one-way: once the file is restructured, re-running `update-project.sh` classifies it as `current` and skips it. The backup in `agent-vault/context/updates/<timestamp>/` is the recovery path.

## Suggested Additional Tests
- None — the four scenarios cover `missing`, `current` (implicit via freshly bootstrapped repos in existing tests), `legacy-known` (real + dry-run), and `unknown`.

PR #95, opened 2026-04-24
Add opt-in metadata-only main push gate

### LINKED ISSUE #94: Allow docs-only direct-push to main for agent-vault metadata refreshes
> 🤖 **Post by Claude Opus 4.7 (1M context)** · via Claude Code

## Problem

Consumer repos that pair the agent-vault workflow with a direct-push-to-main policy rail (e.g. a Bash hook in Claude Code that blocks pushes to \`main\`) end up filing a full PR every time they need to refresh \`agent-vault/\` metadata after a feature PR merges.

A typical post-merge refresh is 3 files: \`agent-vault/context-log.md\` (Current Snapshot + new entry), \`agent-vault/daily/YYYY-MM-DD.md\` (post-merge section), and \`agent-vault/design-log/...md\` (cleanup note). Zero source changes. These refreshes carry no meaningful review value in practice — automated reviewers don't flag docs-only agent-vault changes and human reviewers don't inspect them.

The PR-per-refresh cost is ~5 steps (branch, commit, push, open PR, merge, cleanup) for changes that aren't substantively different from amending a commit message.

## Proposed fix

Ship an opt-in variant of the push-gate / pre-commit hook in \`_assets/\` (or wherever the hook lives) that **allows direct push to \`main\` when every changed path is under \`agent-vault/\`**.

Pseudocode:

\`\`\`bash
changed=\$(git diff --name-only HEAD@{1} HEAD)
if [[ -z \"\$changed\" ]] || echo \"\$changed\" | grep -qvE '^agent-vault/'; then
  # At least one non-agent-vault file touched → enforce PR flow.
  fail_if_pushing_to_main
fi
\`\`\`

Source code, tests, configuration, and scripts all still require a PR. Only the metadata-only subset bypasses the gate.

## Rollback

If a consumer-repo pattern emerges where metadata PRs accidentally carry source changes that slip through, tighten the hook back and treat the scoped variant as opt-out.

## Why this belongs in agent-vault

The hook lives in agent-vault's scaffold surface, and the metadata-shape that justifies the carve-out (\`context-log.md\` + \`daily/\` + \`design-log/\`) is agent-vault's invention. Consumer repos shouldn't have to reinvent the carve-out themselves.

ssheld:
## Summary

Closes #94.

This adds an opt-in tracked `pre-push` gate that lets generated repos allow direct pushes to `main` only when the pushed range contains runtime `agent-vault` metadata. The hook is inert by default and activates only from repo-local config, so existing consumers do not inherit a direct-push policy change unless they opt in.

## Files / Areas Changed

- `scaffold/agent-vault/_assets/hooks/lib/runtime-note.sh`: extracts the runtime metadata path classifier so `pre-commit` and `pre-push` share one allowlist.
- `scaffold/agent-vault/_assets/hooks/pre-commit`: sources the shared classifier instead of carrying a duplicate allowlist.
- `scaffold/agent-vault/_assets/hooks/pre-push`: adds the opt-in main push gate, including local-only config, main-only evaluation, delete/create/non-fast-forward rejection, and conservative `git diff --no-renames --name-only` path checks.
- `scripts/new-project.sh` and `scripts/update-project.sh`: propagate the new hook assets and preserve executable bits for hook files.
- `README.md`, `scaffold/agent-vault/_assets/hooks/README.md`, `scaffold/agent-vault/shared-rules.md`, and mirrored `AGENTS.md`: document the policy line that direct push is for recording runtime history, while behavior-changing files stay on the PR path.
- `.github/workflows/scaffold-regression-checks.yml` and hook tests: add CI coverage for the main push gate.

## Verification

- `bash scripts/test-main-push-gate.sh` - passed
- `bash scripts/test-session-metadata-hook.sh` - passed
- `bash scripts/test-gitignore-management.sh` - passed
- `bash scripts/test-coding-standards-sync.sh` - passed
- `bash scripts/test-decision-template-sync.sh` - passed
- `bash scripts/test-new-worktree.sh` - passed
- `bash scripts/test-remove-worktree.sh` - passed
- `bash scripts/test-worktree-helper-sync.sh` - passed
- `bash -n` on changed shell scripts/hooks - passed
- `git diff --check` - passed

## Self-Review Notes

During self-review, I found one missing regression for the repo-local config requirement. I added coverage proving a global `agent-vault.allowMetadataOnlyMainPush=true` setting is ignored unless the current repo opts in locally.

## Risks / Rollback

- Risk: generated repos that opt in may initially be surprised when behavior-changing `agent-vault` files are blocked from direct `main` pushes. The hook prints the offending paths and points back to the PR flow.
- Risk: the allowlist could drift if future hook logic bypasses the shared classifier. This PR centralizes the classifier and tests blocked policy/template/hook paths to reduce that risk.
- Rollback: unset the opt-in in a consumer repo with `git config --local --unset agent-vault.allowMetadataOnlyMainPush`, or revert the hook asset changes in this template.

## Docs Consistency

No impact on a root `docs/design.md` architecture document was detected. README and scaffold hook/policy docs were updated to describe the new opt-in behavior and intentional exclusions.

### DIFF
--- a/.github/workflows/scaffold-regression-checks.yml
+++ b/.github/workflows/scaffold-regression-checks.yml
@@ -13,6 +13,7 @@ on:
       - "scripts/test-coding-standards-sync.sh"
       - "scripts/test-decision-template-sync.sh"
       - "scripts/test-session-metadata-hook.sh"
+      - "scripts/test-main-push-gate.sh"
       - "scripts/test-new-worktree.sh"
       - "scripts/test-remove-worktree.sh"
       - "scripts/test-worktree-helper-sync.sh"
@@ -31,6 +32,7 @@ on:
       - "scripts/test-coding-standards-sync.sh"
       - "scripts/test-decision-template-sync.sh"
       - "scripts/test-session-metadata-hook.sh"
+      - "scripts/test-main-push-gate.sh"
       - "scripts/test-new-worktree.sh"
       - "scripts/test-remove-worktree.sh"
       - "scripts/test-worktree-helper-sync.sh"
@@ -51,6 +53,8 @@ jobs:
         run: bash scripts/test-decision-template-sync.sh
       - name: Verify session metadata hook behavior
         run: bash scripts/test-session-metadata-hook.sh
+      - name: Verify main push metadata gate
+        run: bash scripts/test-main-push-gate.sh
       - name: Verify worktree creation helper
         run: bash scripts/test-new-worktree.sh
       - name: Verify worktree removal helper
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ Run the scaffold regression scripts locally when changing bootstrap, sync, or tr
 - `bash scripts/test-coding-standards-sync.sh`
 - `bash scripts/test-decision-template-sync.sh`
 - `bash scripts/test-session-metadata-hook.sh`
+- `bash scripts/test-main-push-gate.sh`
 - `bash scripts/test-new-worktree.sh`
 - `bash scripts/test-remove-worktree.sh`
 - `bash scripts/test-worktree-helper-sync.sh`
@@ -111,7 +112,9 @@ CI also runs these checks via `.github/workflows/scaffold-regression-checks.yml`
   - `<repo>/agent-vault/GEMINI.md`
   - `<repo>/agent-vault/handoff.md`
   - `<repo>/agent-vault/_assets/hooks/README.md`
+  - `<repo>/agent-vault/_assets/hooks/lib/runtime-note.sh`
   - `<repo>/agent-vault/_assets/hooks/pre-commit`
+  - `<repo>/agent-vault/_assets/hooks/pre-push`
   - `<repo>/agent-vault/design-log/README.md`
   - `<repo>/agent-vault/context/handoffs/README.md`
   - `<repo>/agent-vault/decisions/README.md`
@@ -177,7 +180,10 @@ When a managed file changes, the script backs up the previous version under:
 Generated projects auto-enable the tracked metadata gate in the clone where `new-project.sh` or `update-project.sh` runs, unless `core.hooksPath` is already set to something else. Additional clones can enable it with:
 - `git -C <repo-path> config core.hooksPath agent-vault/_assets/hooks`
 
-That tracked hook enforces the baseline session artifacts and validates staged `agent-vault/context-log.md` ordering/freshness so newer entries do not get appended below stale headers.
+The tracked `pre-commit` hook enforces the baseline session artifacts and validates staged `agent-vault/context-log.md` ordering/freshness so newer entries do not get appended below stale headers. The tracked `pre-push` hook is inert by default; repos may opt into a narrow direct-push-to-`main` shortcut for runtime `agent-vault` metadata only:
+- `git -C <repo-path> config --local agent-vault.allowMetadataOnlyMainPush true`
+
+That shortcut allows recording history after PR merges while keeping source code, config, scripts, root docs, policy files, templates, hook assets, and durable project docs such as `agent-vault/README.md`, `plan.md`, `coding-standards.md`, `project-context.md`, `project-commands.md`, and `handoff.md` on the PR path.
 When syncing older generated repos, `update-project.sh` now auto-migrates recognized legacy `agent-vault/context-log.md` layouts into the validator-compatible top-level `## Current Snapshot` / `## Entries` shape before syncing the stricter hook. If the layout is not recognized, the script leaves the file unchanged and prints a manual-remediation warning instead of guessing.
 
 Both scripts also ensure root `.gitignore` includes managed local-only ignore entries (added only when missing):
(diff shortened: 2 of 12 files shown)


ssheld (commented):
> 🤖 **Review by Claude Opus 4** · via Claude Code
Formal review state unavailable: GitHub blocks self-approval on your own PR; submitting as a COMMENT review with the textual merge recommendation below.

## Strengths
- Opt-in via `--local` only (global config intentionally ignored) keeps the policy change from leaking into unrelated clones; this is explicitly covered by the `global-config-ignored` test.
- Classifier extraction into `scaffold/agent-vault/_assets/hooks/lib/runtime-note.sh` removes the duplicate allowlist between `pre-commit` and `pre-push`, and the rename-smuggling test (`git mv agent-vault/AGENTS.md agent-vault/daily/...`) verifies that `--no-renames` makes the gate conservative.
- The test matrix in `scripts/test-main-push-gate.sh` is thorough: opt-out, metadata-only, source-blocked, mixed-blocked, non-main-unaffected, per-path blocked set, delete/create/non-FF rejection, and invalid config value.

## Merge Recommendation
Approve

## Top Risks
- Enforcement depends on `core.hooksPath` pointing at `agent-vault/_assets/hooks`. A consumer who sets the opt-in flag but forgets the hooks path will silently have no gate; remote branch protection is still the durable backstop.
- `git config --type=bool` requires git 2.18+ (mid-2018). Consumers on older git would hit an 'unknown option' error on every push once opted in. Very unlikely in practice.
- Allowlist drift remains a latent risk if future runtime-note paths are added (for example a new `agent-vault/<dir>/`); the centralized classifier plus the blocked-path tests mitigate but do not eliminate this.

## Suggested Additional Tests
- Explicit coverage that the hook is a no-op outside an `agent-vault/`-bearing repo (the early `[[ ! -d "agent-vault" ]]` branch at `scaffold/agent-vault/_assets/hooks/pre-push:15` is currently untested).
- A push that updates `main` with zero file changes between `remote_sha` and `local_sha` (for example an empty merge commit or identical tree) to pin the 'no blocked paths, FF-only' pass path.

ssheld:

I reviewed all feedback surfaces for PR #95: top-level review submissions, inline review comments/threads, and PR conversation comments. One inline P1 and two suggested additional tests were actionable.

| 1 | Inline P1 on `scaffold/agent-vault/_assets/hooks/pre-push:66`: endpoint-tree diff could miss blocked files changed and reverted inside `remote_sha..local_sha`. | Resolved | `ef9e7c4`; `scaffold/agent-vault/_assets/hooks/pre-push`, `scripts/test-main-push-gate.sh`; the hook now walks `git rev-list --reverse "$remote_sha..$local_sha"` and classifies changed paths per commit, so intermediate blocked-file commits are rejected even when the final tree is clean. |
| 2 | Claude suggested explicit coverage that the hook is a no-op outside an `agent-vault/` repo. | Resolved | `ef9e7c4`; `scripts/test-main-push-gate.sh`; added `no-agent-vault-noop`, which runs the hook from a plain repo with local opt-in set and verifies it exits cleanly. |
| 3 | Claude suggested coverage for a push with zero file changes between `remote_sha` and `local_sha`. | Resolved | `ef9e7c4`; `scripts/test-main-push-gate.sh`; added `empty-commit-allowed`, which verifies an empty fast-forward commit to `main` passes when no blocked paths are present. |

## Remaining Items
- None from the retrieved review feedback.

## Verification
- `bash scripts/test-main-push-gate.sh` - passed
- `bash scripts/test-session-metadata-hook.sh` - passed
- `bash -n scaffold/agent-vault/_assets/hooks/pre-push` - passed
- `bash -n scripts/test-main-push-gate.sh` - passed
- `git diff --check` - passed

ssheld (commented):
> 🤖 **Review by Claude Opus 4** · via Claude Code
Formal review state unavailable: GitHub blocks self-approval on your own PR; submitting as a COMMENT review.

Re-reviewed at `ef9e7c4`. All prior feedback is resolved and I found no regressions.

| 1 | Codex P1 — endpoint-tree diff could miss files changed+reverted inside `remote_sha..local_sha` | Resolved | `ef9e7c4`; `scaffold/agent-vault/_assets/hooks/pre-push` walks `git rev-list --reverse "$remote_sha..$local_sha"` and classifies per commit; `scripts/test-main-push-gate.sh` adds `intermediate-blocked-file-blocked` |
| 2 | Claude suggested test — hook is a no-op outside an `agent-vault/` repo | Resolved | `ef9e7c4`; `scripts/test-main-push-gate.sh` adds `no-agent-vault-noop` |
| 3 | Claude suggested test — FF push with zero file changes | Resolved | `ef9e7c4`; `scripts/test-main-push-gate.sh` adds `empty-commit-allowed` |

## Regression Check on `ef9e7c4`
- Per-commit walk is a strict superset of the previous endpoint diff: for single-commit pushes, `git rev-list` yields one SHA and `diff-tree`/`diff --no-renames` against its parent equals the former endpoint diff, so the pre-existing source-blocked, mixed-blocked, non-main-unaffected, per-path blocked, and rename tests continue to pass unchanged.
- `--no-renames` is preserved at the inner diff, so the rename-smuggling defense from the original PR still holds.
- Root-commit branch in `collect_commit_changed_files` correctly falls back to `git diff-tree --no-commit-id --name-only -r --no-renames --root` when `"$commit_sha^1"` fails to resolve.
- FF-with-merge-commit case is covered implicitly: any commit reachable from `local_sha` but not `remote_sha` is visited individually by `rev-list`, so files introduced via a merged-in branch are classified on their own commits rather than only via the merge's first-parent diff.

## Strengths
- Fix addresses the root cause (transient history on `main`) rather than masking it — blocked content is rejected even when the endpoint tree is clean.
- New tests cover both directions (`intermediate-blocked-file-blocked` for the P1 scenario, `empty-commit-allowed` for the benign case), which locks in behavior against future refactors.

## Merge Recommendation
Approve

## Top Risks
- Same as prior review: enforcement still requires `core.hooksPath` to be set to `agent-vault/_assets/hooks`; remote branch protection remains the durable backstop.
- Performance scales with commits in the pushed range. Not a concern for the intended metadata-shortcut workflow.

## Suggested Additional Tests
- None — the existing matrix now covers endpoint, intermediate, empty, rename, and no-vault edges.

PR #96, opened 2026-04-24
Fix stale worktree branch cleanup

ssheld:
> 🤖 Prepared by Codex GPT-5 · via Codex CLI

## Summary
- Problem: `scripts/remove-worktree.sh --branch <name> --delete-branch` could not recover when the branch's worktree directory had already been manually deleted. The first run pruned stale worktree metadata and exited; the rerun then failed with `No worktree found for branch`, leaving the local branch stuck.
- Approach: Teach the managed scaffold helper to treat `--delete-branch` without `--path` as an explicit branch-only cleanup request when no live worktree remains, while preserving the existing error for non-delete invocations.
- Outcome: Generated projects can now clean up a lingering local branch after stale worktree metadata is pruned.

## Changes
- `scaffold/root/scripts/remove-worktree.sh`: added `delete_local_branch` and branch-only cleanup paths for missing/stale worktrees when `--delete-branch` is explicitly requested.
- `scripts/test-remove-worktree.sh`: extended the stale-record regression to verify the rerun deletes the lingering branch successfully.

## Verification Loop
- Build / package: Not applicable - shell helper scaffold change only.
- Typecheck / static analysis: Not applicable - no typed language surface changed.
- Lint / format validation: `bash -n scaffold/root/scripts/remove-worktree.sh`; `bash -n scripts/test-remove-worktree.sh`; `git diff --check` - passed.
- Tests / coverage: `bash scripts/test-remove-worktree.sh` - 38 passed, 0 failed. `bash scripts/test-worktree-helper-sync.sh` - 27 passed, 0 failed.
- Security / secrets / workflow checks: Reviewed the shell path handling. Branch deletion remains opt-in via `--delete-branch`; path-based cleanup still refuses missing paths and branch/path mismatches.
- Diff review: Reviewed final diff before commit `65a970a`.
- Not run: Full repository test sweep was not run because the change is isolated to the worktree cleanup helper and its sync behavior; targeted helper regressions covered the changed behavior.

## Risks and Rollback
- Risk: Users who pass `--branch <name> --delete-branch` after a stale prune now delete the branch immediately instead of receiving another error.
- Mitigation: This only happens when `--delete-branch` is explicit and no `--path` is supplied, matching the documented cleanup intent.
- Rollback: Revert `65a970a` to restore the previous stale-record behavior.

## Docs Consistency
- `docs/design.md`: No impact - helper script behavior only.
- `README.md`: No impact - existing cleanup command remains correct; behavior now matches that command in the stale-record case.

## Follow-Ups
- [ ] None

### DIFF
--- a/scaffold/root/scripts/remove-worktree.sh
+++ b/scaffold/root/scripts/remove-worktree.sh
@@ -35,6 +35,14 @@ die() {
     exit 1
 }
 
+delete_local_branch() {
+    local branch_name="$1"
+
+    git -C "$PROJECT_DIR" show-ref --verify --quiet "refs/heads/$branch_name" \
+        || die "No local branch found: $branch_name"
+    git -C "$PROJECT_DIR" branch -D "$branch_name"
+}
+
 resolve_path() {
     local raw="$1"
     if [[ "$raw" = /* ]]; then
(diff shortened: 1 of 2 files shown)


ssheld (commented):
> 🤖 **Review by Claude Opus 4** · via Claude Code
Formal review state unavailable: PR is already merged; posting retrospective feedback as a review-mode COMMENT.

No findings above the confidence threshold. One optional test-coverage note below.

## Strengths
- Two-branch auto-cleanup matches user intent: both the "record already pruned" path and the "record still references a missing directory" path now terminate successfully when `--delete-branch` is explicit and `--path` is absent. The `[[ -z "$TARGET_PATH" ]]` guard correctly preserves the old behavior when the caller disambiguated with a path.
- The updated die message (`pass --delete-branch without --path to delete the remaining local branch`) is strictly more actionable than the previous `rerun this command to remove the remaining branch`, since it tells the user exactly which flag enables the new path.
- Extracting `delete_local_branch` and routing the tail `--delete-branch` path through it collapses three near-duplicate call sites and adds a `show-ref --verify` precheck, so the error message is consistent if the branch is already gone.

## Merge Recommendation
Approve (post-merge acknowledgement)

## Top Risks
- `delete_local_branch` uses `git branch -D` (force delete), matching the pre-existing tail call. For an orphaned worktree whose commits were never pushed, this silently discards unmerged history. Not a regression from this PR, but the flag name `--delete-branch` does not hint at force semantics; worth revisiting if user reports confusion.
- Branch-only cleanup only triggers when `--branch` was passed; a caller who specified `--path` against a now-missing worktree still dies without auto-deleting. This is intentional (path was explicit, so don't reinterpret intent), but worth documenting in the helper's `--help` if future users expect symmetry.

## Suggested Additional Tests
- The new assertion covers **path 1** (record already pruned → `--delete-branch` succeeds). **Path 2** (record still present, directory missing, first-time `--delete-branch` invocation → prune + branch delete in one shot) is reachable from line 212 of the updated helper but not exercised by a dedicated assertion. A one-shot stale + `--delete-branch` variant of Test 10 would pin the "Pruned stale worktree record for missing directory" message path and prevent regressions that re-introduce the original two-run workflow.

PR #97, opened 2026-04-25
Harden agent-vault hook deletion checks

ssheld:
## Summary
- Problem: CritterCam PR feedback identified that generated hooks could no-op when `agent-vault/` was missing from the working tree, which weakens deletion-case enforcement for the optional metadata-only push gate.
- Approach: Fix the scaffold source hooks in `agent-vault` instead of patching CritterCam directly, and add regression coverage for missing-vault deletion scenarios.
- Outcome: The pre-push main gate now rejects enabled `main` pushes when the generated `agent-vault/` directory or runtime classifier is missing, and the pre-commit hook rejects staged `agent-vault/` path changes when the directory has been removed.

## Changes
- `scaffold/agent-vault/_assets/hooks/pre-push`: defer loading the runtime metadata classifier until a `main` update needs it, and reject `main` pushes when `agent-vault/` or the classifier is missing while `agent-vault.allowMetadataOnlyMainPush` is enabled.
- `scaffold/agent-vault/_assets/hooks/pre-commit`: collect staged files before checking for `agent-vault/`, no-op for plain repos with no staged vault paths, and reject staged `agent-vault/` changes when the directory is missing.
- `scripts/test-main-push-gate.sh`: cover the enabled missing-vault `main` push rejection while preserving plain-repo no-op behavior when the gate is not locally enabled.
- `scripts/test-session-metadata-hook.sh`: cover the staged `agent-vault/` deletion rejection path for the pre-commit hook.

## Validation
- `bash -n scaffold/agent-vault/_assets/hooks/pre-commit`: passed.
- `bash -n scaffold/agent-vault/_assets/hooks/pre-push`: passed.
- `bash -n scripts/test-main-push-gate.sh`: passed.
- `bash -n scripts/test-session-metadata-hook.sh`: passed.
- `git diff --check`: passed.
- `git diff --cached --check`: passed before commit.
- `bash scripts/test-main-push-gate.sh`: passed.
- `bash scripts/test-session-metadata-hook.sh`: passed.

## Risks and Rollback
- Risk: Repos with the optional metadata-only main-push gate enabled and a missing generated `agent-vault/` directory will now fail direct `main` pushes instead of silently no-oping.
- Mitigation: This is intentional for inconsistent generated-repo state; plain repos without the local opt-in still no-op, and non-main pushes are unaffected by the pre-push consistency check.
- Rollback: Revert commit `966a66c`.

## CritterCam Follow-Up
- Do not rerun `scripts/update-project.sh` into CritterCam until this PR merges.
- After merge, update the local `agent-vault` checkout to merged `main`, rerun the normal update script against `workspaces/crittercam`, then update CritterCam PR `#331` and respond to its feedback with the new commit evidence.

### DIFF
--- a/scaffold/agent-vault/_assets/hooks/README.md
+++ b/scaffold/agent-vault/_assets/hooks/README.md
@@ -17,6 +17,8 @@ git config core.hooksPath agent-vault/_assets/hooks
     - `agent-vault/context-log.md`
     - one note under `agent-vault/daily/`
     - one note under `agent-vault/design-log/`
+  - Rejects staged `agent-vault/` path changes when the generated `agent-vault/`
+    directory or shared runtime metadata classifier is missing.
   - Validates staged `agent-vault/context-log.md` content:
     - entry headings must use `YYYY-MM-DD HH:MM local - <agent> - <topic>`
     - entries must remain newest-first
@@ -26,6 +28,8 @@ git config core.hooksPath agent-vault/_assets/hooks
   - Inert by default.
   - When explicitly enabled with local repo config, blocks direct pushes to `main` unless every pushed path is runtime `agent-vault` metadata.
   - Rejects direct deletion of `main`, first-time creation of `main`, and non-fast-forward pushes.
+  - Rejects `main` pushes when the generated `agent-vault/` directory or shared
+    runtime metadata classifier is missing.
   - Uses the same runtime metadata classifier as `pre-commit`.
 
 ## Optional Direct Push to Main for Runtime Metadata
--- a/scaffold/agent-vault/_assets/hooks/pre-commit
+++ b/scaffold/agent-vault/_assets/hooks/pre-commit
@@ -13,8 +13,32 @@ fi
 
 cd "$repo_root"
 
-if [[ ! -d "agent-vault" ]]; then
+all_staged_files=()
+while IFS= read -r file_path; do
+  [[ -n "$file_path" ]] || continue
+  all_staged_files+=("$file_path")
+done < <(git diff --cached --name-only --diff-filter=ACDMRTUXB)
+
+if [[ "${#all_staged_files[@]}" -eq 0 ]]; then
   exit 0
+fi
+
+if [[ ! -d "agent-vault" ]]; then
+  for file_path in "${all_staged_files[@]}"; do
+    if [[ "$file_path" == "agent-vault" || "$file_path" == agent-vault/* ]]; then
+      echo "agent-vault metadata gate failed." >&2
+      echo "Tracked agent-vault directory is missing while staged changes include agent-vault paths." >&2
+      exit 1
+    fi
+  done
+
+  exit 0
+fi
+
+if [[ ! -f "agent-vault/_assets/hooks/lib/runtime-note.sh" ]]; then
+  echo "agent-vault metadata gate failed." >&2
+  echo "Runtime metadata classifier is missing." >&2
+  exit 1
 fi
 
 # shellcheck source=./lib/runtime-note.sh
@@ -157,16 +181,6 @@ validate_staged_context_log() {
   printf '%s\n' "${errors[@]}"
   return 1
 }
-
-all_staged_files=()
-while IFS= read -r file_path; do
-  [[ -n "$file_path" ]] || continue
-  all_staged_files+=("$file_path")
-done < <(git diff --cached --name-only --diff-filter=ACDMRTUXB)
-
-if [[ "${#all_staged_files[@]}" -eq 0 ]]; then
-  exit 0
-fi
 
 non_deleted_staged_files=()
 while IFS= read -r file_path; do
--- a/scaffold/agent-vault/_assets/hooks/pre-push
+++ b/scaffold/agent-vault/_assets/hooks/pre-push
@@ -12,10 +12,6 @@ fi
 
 cd "$repo_root"
 
-if [[ ! -d "agent-vault" ]]; then
-  exit 0
-fi
-
 config_value=""
 config_rc=0
 config_value="$(git config --local --get --type=bool "$config_key" 2>/dev/null)" || config_rc=$?
@@ -29,9 +25,6 @@ if [[ "$config_value" != "true" ]]; then
   exit 0
 fi
 
-# shellcheck source=./lib/runtime-note.sh
-source "agent-vault/_assets/hooks/lib/runtime-note.sh"
-
 reject_main_push() {
   local reason="$1"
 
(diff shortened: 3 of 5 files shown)


ssheld (commented):

1. `scaffold/agent-vault/_assets/hooks/pre-push:48`, `scaffold/agent-vault/_assets/hooks/pre-commit:38`

   **Optional** · High confidence — Missing test coverage for the new "classifier file missing" rejection branch.
   Both hooks now have a dedicated check for `agent-vault/_assets/hooks/lib/runtime-note.sh` being missing while the rest of the gate is in scope (`pre-push:48-50`, `pre-commit:38-42`). The new tests in `scripts/test-main-push-gate.sh` and `scripts/test-session-metadata-hook.sh` cover the directory-missing case but not the "directory present, lib file deleted" case. A small fixture that removes only `agent-vault/_assets/hooks/lib/runtime-note.sh` would lock that branch in. Without it, a future regression that breaks the classifier-existence check would silently pass tests as long as the directory check still works.

2. `scaffold/agent-vault/_assets/hooks/README.md:24-28`

   **Optional** · Medium confidence — Hook README contract widened, but the README is unchanged.
   The current `pre-push` bullets describe inert-by-default plus the deletion/creation/non-fast-forward rejections, and `pre-commit` describes the staged-metadata gate. Neither mentions the new "rejects when `agent-vault/` or `lib/runtime-note.sh` is missing while the gate is in scope" branch. A consumer reading the README to understand what the hook will block would miss this. Consider adding one bullet to each hook section so the documented behavior matches what the tests now enforce. Downstream consumers (e.g., CritterCam) will sync this README on the next scaffold update, so any wording added here propagates automatically.

## Strengths
- Directly addresses the `pre-push` directory-presence bypass surfaced on the consumer side (CritterCam PR `#331`) and fixes it in the scaffold source instead of patching the consumer, which keeps every downstream sync aligned. Same fix applied symmetrically to `pre-commit`.
- The `pre-push` fix is surgical: non-`main` refs still skip the loop body via `[[ "$remote_ref" == "refs/heads/main" ]] || continue`, so the new directory/classifier checks only fire on the case that matters. Lazy-loading the classifier through `ensure_runtime_note_classifier` with a `runtime_note_classifier_loaded` flag avoids redundant sourcing across multiple `main`-targeted ref updates in a single push and keeps non-main pushes from paying any cost.
- Test suite is updated alongside the behavior change. The `pre-commit` test correctly seeds with `AGENT_VAULT_SKIP_METADATA_GATE=1` so the bootstrap commit isn't tangled with the gate under test, and the `pre-push` test uses `core.hooksPath=/dev/null` for the same reason during the deletion commit. The validation section also runs the actual functional tests, not just `bash -n`.

## Merge Recommendation
Approve

## Top Risks
- Behavioral change called out in the PR body: repos that had `agent-vault.allowMetadataOnlyMainPush=true` set on a host that does not actually use `agent-vault/` will now fail direct `main` pushes instead of silently no-oping. This is intentional and the right default, but worth flagging in the next downstream sync's release note so consumers don't read it as a regression.
- The `pre-commit` reordering moves `all_staged_files` collection above the function definitions and before the `agent-vault` directory check. Any future edit that re-introduces an early exit or adds a new check between `cd "$repo_root"` and the `all_staged_files` collection needs to preserve the "collect first, then classify" invariant that this PR establishes; otherwise the `agent-vault`-staged-paths-but-missing-directory rejection silently regresses.

## Suggested Additional Tests
- Add a fixture that creates a fully seeded project, deletes only `agent-vault/_assets/hooks/lib/runtime-note.sh` (leaving the directory in place), then exercises both hooks: `pre-push` should reject with the "Runtime metadata classifier is missing" message when the gate is enabled, and `pre-commit` should reject with the same message when staged changes are non-runtime-metadata.
- Optional: add a `pre-push` test for a multi-ref push where one ref is `main` and one is a feature branch, asserting the feature-branch ref is unaffected by the new directory/classifier checks (current loop structure should pass this, but locking it in protects against future refactors that hoist the checks above the `[[ "$remote_ref" == "refs/heads/main" ]] || continue` guard).

ssheld (commented):
> 🤖 **Rereview by Claude Opus 4.7** · via Claude Code

Both findings from my prior review (https://github.com/ssheld/agent-vault/pull/97#pullrequestreview-4175707804) are resolved at `cb3796f`. The optional multi-ref push test from "Suggested Additional Tests" is also implemented. No regressions detected.

## Prior Findings Status

| 1 | Missing test coverage for the classifier-file-missing rejection branch (`pre-push:48-50`, `pre-commit:38-42`) | Resolved | `scripts/test-main-push-gate.sh` adds `missing_classifier_repo` fixture that removes only `agent-vault/_assets/hooks/lib/runtime-note.sh` and asserts pre-push rejects with `Runtime metadata classifier is missing while agent-vault.allowMetadataOnlyMainPush is enabled.`; `scripts/test-session-metadata-hook.sh` adds the matching pre-commit fixture asserting `Runtime metadata classifier is missing.` |
| 2 | Hook README contract widened but README unchanged (`scaffold/agent-vault/_assets/hooks/README.md:24-28`) | Resolved | Two new bullets at `scaffold/agent-vault/_assets/hooks/README.md:20-21` (pre-commit) and `scaffold/agent-vault/_assets/hooks/README.md:31-32` (pre-push) documenting `Rejects ... when the generated agent-vault/ directory or shared runtime metadata classifier is missing.` Wording is symmetric across both hook sections. |
| ST2 (optional suggestion) | Multi-ref push test mixing `main` with a feature branch | Implemented | `scripts/test-main-push-gate.sh` adds `multi_ref_repo` fixture that pushes both `refs/heads/feature-side` and `refs/heads/main` together with the gate enabled; expected to succeed because the feature-side ref skips the main-loop body via the existing `refs/heads/main` guard, and the main ref's update is metadata-only. |

## Regression Check

- `pre-commit` "collect first, then classify" invariant preserved: `all_staged_files` is collected at `scaffold/agent-vault/_assets/hooks/pre-commit:16-24`, the directory-missing rejection runs at `:26-36`, and the classifier-file rejection runs at `:38-42`. The downstream `non_deleted_staged_files` collection and `case "$file_path"` walks still see the populated arrays.
- `pre-push` lazy-load preserved: `ensure_runtime_note_classifier` is invoked at `scaffold/agent-vault/_assets/hooks/pre-push:81`, inside the main-ref loop body, after the deletion / creation / non-fast-forward rejection checks. The `[[ "$remote_ref" == "refs/heads/main" ]] || continue` guard at `scaffold/agent-vault/_assets/hooks/pre-push:70` is unchanged, so non-main refs still bypass the new directory/classifier checks. The `runtime_note_classifier_loaded` idempotency flag at `:37` keeps multi-ref main pushes from re-sourcing the classifier.
- `scripts/test-main-push-gate.sh` plain `no_vault_repo` case at `:138-146` no longer enables the gate (line removed) and still asserts plain-repo no-op behavior. The "gate enabled + missing vault" case is now intentionally a rejection, covered by the new `missing_vault_repo` fixture.
- PR Validation section reruns `bash scripts/test-main-push-gate.sh` and `bash scripts/test-session-metadata-hook.sh` and reports both passing.

## Strengths

- Resolved both prior findings in one follow-up commit and implemented the optional multi-ref suggestion. Test coverage now locks in directory-missing and classifier-only-missing branches symmetrically across pre-commit and pre-push.
- Multi-ref push test guards against future refactors that hoist the directory/classifier checks above the `refs/heads/main` guard, which would silently make feature-branch pushes fail on enabled-gate repos with no vault.
- README bullets use parallel wording for both hooks, so the documented behavior matches the test assertions and propagates cleanly to downstream consumers (e.g., CritterCam) on the next scaffold update.

## Merge Recommendation

Approve

## Top Risks

- Same intentional behavioral change called out in the PR body: repos with `agent-vault.allowMetadataOnlyMainPush=true` and a missing or partially-missing `agent-vault/` will now fail direct `main` pushes instead of silently no-oping. Worth a release note when this propagates downstream so consumers don't read it as a regression.

## Suggested Additional Tests

- None.

Issue #99, opened 2026-04-26
Add lightweight style checks for shell scripts


ssheld:
> 🤖 **Post by Codex GPT-5** · via Codex CLI

## Problem

The repository has several shell scripts and generated shell helper templates, but no dedicated style/static check command. Regression scripts catch behavior, but they do not consistently catch shell footguns, formatting drift, or syntax issues before review.

## Proposed Direction

Add a lightweight style-check path focused on shell first, without introducing a heavy package-manager/toolchain dependency just for linting.

## Suggested Scope

- Add a repo command such as `scripts/check-style.sh`.
- Include shell checks for relevant tracked scripts:
  - `bash -n` syntax validation.
  - `shellcheck` for shell static analysis.
  - `shfmt -d` for shell formatting drift.
- Cover at least:
  - `scripts/*.sh`
  - `scripts/lib/*.sh`
  - `scaffold/root/scripts/*.sh`
  - generated hook assets under `scaffold/agent-vault/_assets/hooks/` where applicable.
- Document the command in `README.md`.
- Add CI coverage so reviewers do not need local tool installs to catch style/static issues.

## Out Of Scope For First Pass

- Broad Markdown linting. It may be useful later, but only with a low-noise config that avoids generic line-length/list churn.
- Adding a heavyweight JS/Python toolchain solely to run style checks.
- Reformatting unrelated files.

## Acceptance Criteria

- A single local command runs the style/static checks.
- CI runs the same command.
- Missing local optional tools fail with actionable install guidance or are documented clearly.
- Existing regression scripts remain separate from style checks.
- Documentation explains when to run the command.

ssheld:
> 🤖 **Post by Codex GPT-5** · via Codex CLI

## Implementation Plan

1. Add `scripts/check-style.sh` as the single local entrypoint.
   - Use `set -euo pipefail` and resolve the repo root from the script location.
   - Build the target list from tracked files rather than raw shell globs, so CI and local runs behave the same.
   - Cover the current shell surface:
     - `scripts/*.sh`
     - `scripts/lib/*.sh`
     - `scaffold/root/scripts/*.sh`
     - `scaffold/agent-vault/_assets/hooks/pre-commit`
     - `scaffold/agent-vault/_assets/hooks/pre-push`
     - `scaffold/agent-vault/_assets/hooks/lib/*.sh`
[...]

PR #101, opened 2026-04-26
Add shell style checks

ssheld:
Refs #99

## Summary

Adds a lightweight shell style/static check path for the template repository. The new local command discovers the tracked shell surface, runs syntax checks, ShellCheck, and shfmt drift checks, and keeps regression checks separate from style checks.

## Changed Areas

- `scripts/check-style.sh`: new single entrypoint for shell syntax, ShellCheck, and shfmt checks, including explicit target discovery, shebang-based hook detection, `--fix` mode, and missing-tool guidance.
- `.github/workflows/style-check.yml`: new dedicated CI workflow that installs pinned ShellCheck `v0.11.0` and shfmt `v3.13.1`, verifies both downloads with SHA-256 digests, then runs the same local command.
- `README.md`: documents when to run the command, local prerequisites for macOS and Ubuntu, CI pins, checksum verification, and the warning-level ShellCheck/shfmt policy.
- Shell scripts and hook assets: applied `shfmt -i 2 -ci` cleanup and fixed warning-level ShellCheck findings needed for the new check to pass.
- `scripts/test-path-expansion.sh`: adds regression coverage for `new-project.sh` and `update-project.sh` repo path expansion for `~`, `~/...`, `/~/...`, and absolute paths.
- `scripts/test-session-metadata-hook.sh`: keeps the new dry-run output assertion that consumes and validates the previously unused `dryrun_output` value.

## Review Feedback Addressed

- Added explicit regression coverage for the `expand_path` fix so literal `~/...` inputs do not regress to `$HOME/~/...`.
- Kept and documented the dry-run message assertion as a real behavior check rather than treating it as formatting-only cleanup.
- Added checksum verification for downloaded CI tool binaries.
- Added concrete Ubuntu local install guidance for ShellCheck and shfmt.

## Verification Loop

Passed:

- `SHELLCHECK_BIN=/tmp/agent-vault-style-tools/shellcheck-v0.11.0/shellcheck SHFMT_BIN=/tmp/agent-vault-style-tools/shfmt_v3.13.1_darwin_arm64 bash scripts/check-style.sh`
- `bash scripts/check-policy-mirrors.sh`
- `bash scripts/test-gitignore-management.sh`
- `bash scripts/test-coding-standards-sync.sh`
- `bash scripts/test-decision-template-sync.sh`
- `bash scripts/test-main-push-gate.sh`
- `bash scripts/test-path-expansion.sh`
- `bash scripts/test-new-worktree.sh`
- `bash scripts/test-remove-worktree.sh`
- `bash scripts/test-worktree-helper-sync.sh`
- `git diff --check`
- `git diff --cached --check`

Also verified that `bash scripts/check-style.sh` without local tools fails with actionable install guidance when `shellcheck` and `shfmt` are not installed on PATH.

## Risks and Rollback

- Main risk is review noise from the first formatting pass, especially in the scaffold worktree helper scripts. The cleanup is limited to the shell target set and covered by existing regression tests.
- CI uses pinned tool versions and pinned digests so future rule, formatting, or binary changes require an intentional workflow update.
- Rollback is straightforward: revert the PR to remove the workflow, local command, documentation, regression test, and associated formatting cleanup.

## Docs Consistency

`README.md` was updated. No `docs/design.md` impact.

### DIFF
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -1,0 +1,61 @@
+name: Style Check
+
+on:
+  pull_request:
+    paths:
+      - "README.md"
+      - "scripts/**"
+      - "scaffold/root/scripts/**"
+      - "scaffold/agent-vault/_assets/hooks/**"
+      - ".github/workflows/style-check.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "README.md"
+      - "scripts/**"
+      - "scaffold/root/scripts/**"
+      - "scaffold/agent-vault/_assets/hooks/**"
+      - ".github/workflows/style-check.yml"
+  workflow_dispatch:
+
+jobs:
+  shell-style:
+    runs-on: ubuntu-latest
+    env:
+      # Pinned so ShellCheck rule changes and shfmt formatting changes are
+      # intentional maintenance bumps rather than runner-image drift.
+      SHELLCHECK_VERSION: v0.11.0
+      SHFMT_VERSION: v3.13.1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install pinned shell style tools
+        run: |
+          set -euo pipefail
+
+          tools_dir="$RUNNER_TEMP/shell-style-tools"
+          bin_dir="$tools_dir/bin"
+          mkdir -p "$bin_dir"
+
+          curl -fsSL \
+            -o "$tools_dir/shellcheck.tar.xz" \
+            "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          tar -C "$tools_dir" -xf "$tools_dir/shellcheck.tar.xz"
+          install -m 0755 "$tools_dir/shellcheck-${SHELLCHECK_VERSION}/shellcheck" "$bin_dir/shellcheck"
+
+          curl -fsSL \
+            -o "$bin_dir/shfmt" \
+            "https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_linux_amd64"
+          chmod +x "$bin_dir/shfmt"
+
+          echo "$bin_dir" >> "$GITHUB_PATH"
+
+      - name: Show tool versions
+        run: |
+          shellcheck --version
+          shfmt --version
+
+      - name: Run shell style checks
+        run: bash scripts/check-style.sh
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -27,6 +27,8 @@ jobs:
       # intentional maintenance bumps rather than runner-image drift.
       SHELLCHECK_VERSION: v0.11.0
       SHFMT_VERSION: v3.13.1
+      SHELLCHECK_SHA256: 8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198
+      SHFMT_SHA256: fb096c5d1ac6beabbdbaa2874d025badb03ee07929f0c9ff67563ce8c75398b1
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -42,12 +44,14 @@ jobs:
           curl -fsSL \
             -o "$tools_dir/shellcheck.tar.xz" \
             "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          printf '%s  %s\n' "$SHELLCHECK_SHA256" "$tools_dir/shellcheck.tar.xz" | sha256sum -c -
           tar -C "$tools_dir" -xf "$tools_dir/shellcheck.tar.xz"
           install -m 0755 "$tools_dir/shellcheck-${SHELLCHECK_VERSION}/shellcheck" "$bin_dir/shellcheck"
 
           curl -fsSL \
             -o "$bin_dir/shfmt" \
             "https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_linux_amd64"
+          printf '%s  %s\n' "$SHFMT_SHA256" "$bin_dir/shfmt" | sha256sum -c -
           chmod +x "$bin_dir/shfmt"
 
           echo "$bin_dir" >> "$GITHUB_PATH"
--- a/scripts/check-style.sh
+++ b/scripts/check-style.sh
@@ -59,7 +59,10 @@ if [[ "${#missing_tools[@]}" -gt 0 ]]; then
 
 Install them locally before running this command:
   macOS:   brew install shellcheck shfmt
-  Ubuntu: install ShellCheck and shfmt from their pinned release binaries.
+  Ubuntu: sudo apt-get update && sudo apt-get install -y shellcheck
+          curl -fsSL -o shfmt https://github.com/mvdan/sh/releases/download/v3.13.1/shfmt_v3.13.1_linux_amd64
+          mkdir -p "$HOME/.local/bin"
+          install -m 0755 shfmt "$HOME/.local/bin/shfmt"
 
 CI pins the exact tool versions in .github/workflows/style-check.yml.
 EOF
--- a/scripts/check-style.sh
+++ b/scripts/check-style.sh
@@ -1,0 +1,159 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SHELLCHECK_BIN="${SHELLCHECK_BIN:-shellcheck}"
+SHFMT_BIN="${SHFMT_BIN:-shfmt}"
+SHELLCHECK_SEVERITY="warning"
+SHFMT_FLAGS=(-i 2 -ci)
+
+usage() {
+  cat <<EOF
+Usage: $0 [--fix|--write]
+
+Run shell syntax, static-analysis, and formatting checks.
+
+Options:
+  --fix, --write  Rewrite shell formatting with shfmt instead of printing a diff.
+  -h, --help      Show this help.
+EOF
+}
+
+die() {
+  echo "Error: $*" >&2
+  exit 1
+}
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+repo_root="$(cd "$script_dir/.." && pwd -P)"
+fix_mode="false"
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --fix | --write)
+      fix_mode="true"
+      shift
+      ;;
+    -h | --help)
+      usage
+      exit 0
+      ;;
+    *)
+      die "Unknown option: $1"
+      ;;
+  esac
+done
+
+cd "$repo_root"
+
+missing_tools=()
+for tool in "$SHELLCHECK_BIN" "$SHFMT_BIN"; do
+  if ! command -v "$tool" >/dev/null 2>&1; then
+    missing_tools+=("$tool")
+  fi
+done
+
+if [[ "${#missing_tools[@]}" -gt 0 ]]; then
+  echo "Missing required style-check tool(s): ${missing_tools[*]}" >&2
+  cat >&2 <<'EOF'
+
+Install them locally before running this command:
+  macOS:   brew install shellcheck shfmt
+  Ubuntu: install ShellCheck and shfmt from their pinned release binaries.
+
+CI pins the exact tool versions in .github/workflows/style-check.yml.
+EOF
+  exit 127
+fi
+
+targets=()
+
+add_target() {
+  local file_path="$1"
+  local existing
+
+  if [[ "${#targets[@]}" -gt 0 ]]; then
+    for existing in "${targets[@]}"; do
+      if [[ "$existing" == "$file_path" ]]; then
+        return 0
+      fi
+    done
+  fi
+
+  targets+=("$file_path")
+}
+
+is_direct_child_sh_file() {
+  local file_path="$1"
+  local directory="$2"
+  local relative_path=""
+
+  case "$file_path" in
+    "$directory"/*.sh)
+      relative_path="${file_path#${directory}/}"
+      [[ "$relative_path" != */* ]]
+      return
+      ;;
+    *)
+      return 1
+      ;;
+  esac
+}
+
+has_shell_shebang() {
+  local file_path="$1"
+  local first_line=""
+
+  IFS= read -r first_line <"$repo_root/$file_path" || return 1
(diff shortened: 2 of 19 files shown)


scripts/new-project.sh
```diff
@@ -14,15 +14,15 @@ usage() {
...
   local p="$1"
   case "$p" in
-    "~")
+    \~)
       printf '%s\n' "$HOME"
       ;;
-    "~/"*)
-      printf '%s/%s\n' "$HOME" "${p#~/}"
+    \~/*)
+      printf '%s/%s\n' "$HOME" "${p#\~/}"
```
ssheld, line 21: 🤖 `Claude Opus 4.7`

**Recommended** · High confidence — Hidden bug fix, not called out in the PR.

This change (`${p#~/}` → `${p#\~/}`, mirrored at `scripts/update-project.sh:28`) silently fixes a real bug. In the original, `~/` inside the strip pattern undergoes tilde expansion to `$HOME/`, which doesn't match the literal `~/` prefix, so the strip is a no-op and the printf produces `$HOME/~/foo` instead of `$HOME/foo`.

Verified empirically with `HOME=/home/test`, `p="~/foo"`:
```
ORIG  ${p#~/}   -> ~/foo      → printf gives /home/test/~/foo  (broken)
NEW   ${p#\~/}  -> foo        → printf gives /home/test/foo    (correct)
```

The `~)` and `/~/` branches are unaffected (no tilde expansion in those positions). This is good — the linter caught a latent bug — but the PR description frames it as `shfmt -i 2 -ci` cleanup, which understates what changed.

Suggest one of:
1. Split into its own commit with a clear `fix:` message.
2. Call it out in the PR body so reviewers know to look here.
3. Add a regression test for `expand_path` covering `~`, `~/foo`, `/~/foo`, and an absolute path so this can't silently re-regress.

scripts/test-session-metadata-hook.sh
```diff
@@ -471,6 +471,7 @@ if [[ "$dryrun_rc" -ne 0 ]]; then
   echo "Expected configure_tracked_hooks_path to return zero on dry-run activation." >&2
   exit 1
 fi
+assert_output_contains "$dryrun_output" "Dry run: would enable tracked metadata hook via core.hooksPath=agent-vault/_assets/hooks"
```
ssheld, line 474: 🤖 `Claude Opus 4.7`

**Recommended** · Medium-High confidence — Real test improvement mislabeled as style cleanup.

This assertion is a new behavior check, not style/format work. Most likely added to silence SC2034 on `dryrun_output` (previously assigned but never read) by giving it a real consumer.

Validating the dry-run message is the better fix versus renaming to `_dryrun_output`, but it expands PR scope beyond what the description claims (`shfmt cleanup and warning-level ShellCheck findings`). Worth either flagging in the PR body or moving to a follow-up so the diff matches its stated intent.

.github/workflows/style-check.yml
```diff
@@ -0,0 +1,61 @@
...
+          curl -fsSL \
+            -o "$tools_dir/shellcheck.tar.xz" \
+            "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          tar -C "$tools_dir" -xf "$tools_dir/shellcheck.tar.xz"
+          install -m 0755 "$tools_dir/shellcheck-${SHELLCHECK_VERSION}/shellcheck" "$bin_dir/shellcheck"
+
+          curl -fsSL \
+            -o "$bin_dir/shfmt" \
+            "https://github.com/mvdan/sh/releases/download/${SHFMT_VERSION}/shfmt_${SHFMT_VERSION}_linux_amd64"
+          chmod +x "$bin_dir/shfmt"
```
ssheld, line 55: 🤖 `Claude Opus 4.7`

**Optional** · Medium-High confidence — Tool downloads lack integrity verification.

The workflow pins ShellCheck `v0.11.0` and shfmt `v3.13.1` by version, but the downloaded artifacts aren't verified. If a release asset on github.com were ever replaced or tampered with, CI would silently install and execute a modified binary that has access to the runner environment and the workflow's `GITHUB_TOKEN`.

Pinning by version is the correct first step. The standard hardening is a `sha256sum -c` against pinned digests for both binaries before installing. Not a blocker, but a small supply-chain seam worth closing on a security-sensitive surface.

scripts/check-style.sh
```diff
@@ -0,0 +1,159 @@
...
+  fi
+done
+
+if [[ "${#missing_tools[@]}" -gt 0 ]]; then
+  echo "Missing required style-check tool(s): ${missing_tools[*]}" >&2
+  cat >&2 <<'EOF'
+
+Install them locally before running this command:
+  macOS:   brew install shellcheck shfmt
+  Ubuntu: install ShellCheck and shfmt from their pinned release binaries.
```
ssheld: 🤖 `Claude Opus 4.7`

**Optional** · Medium confidence — Ubuntu install guidance is vaguer than the plan promised.

The original plan called for "concise install examples for common local environments, such as Homebrew on macOS and apt plus a pinned `shfmt` release on Ubuntu." Currently this message and the README only show the macOS/Homebrew command; the Ubuntu line just says "install ShellCheck and shfmt from their pinned release binaries" without a concrete command.

A concrete `apt-get install shellcheck` line plus the shfmt release URL pattern (matching what CI uses at `style-check.yml:48-51`) would make this actionable for Linux contributors without forcing them to dig through the workflow file.

ssheld (commented):

The implementation matches the revised plan from #99: single entrypoint, dedicated workflow, shebang-based hook detection, `-x` source-following, `--severity=warning`, pinned tool versions with intentional-bump comments, `--fix` mode, and bounded cleanup. The defensive `is_direct_child_sh_file` logic at `scripts/check-style.sh:86` correctly compensates for bash case `*` matching `/`. Path filters and README documentation align with what the plan promised.

Two findings deserve attention because they reach beyond what the PR title and description claim:

1. `scripts/new-project.sh:21` and `scripts/update-project.sh:28`

   **Recommended** · High confidence — Hidden bug fix, not called out anywhere.
   The change `${p#~/}` → `${p#\~/}` silently fixes a real bug in `expand_path`. In the original, `~/` inside `${p#~/}` undergoes tilde expansion (yielding `/home/user/`), which fails to match the literal `~/` in the input, so the strip is a no-op and the printf produces `$HOME/~/foo` instead of `$HOME/foo`. Verified empirically:
   ```
   ORIG ~/foo  -> /home/test/~/foo   # broken
   NEW  ~/foo  -> /home/test/foo     # correct
   ```
   The `~)` and `/~/` branches are unaffected (no expansion in those positions). This is good news — the linter caught a latent bug — but it's framed as "shfmt cleanup and ShellCheck warning fixes," which understates the change. Recommend either splitting it into its own commit with a clear message, calling it out in the PR body, or (best) adding a regression test for `expand_path` covering `~`, `~/foo`, and `/~/foo` so this can't silently re-regress.

2. `scripts/test-session-metadata-hook.sh:474`

   **Recommended** · Medium-High confidence — Test scope creep mislabeled as cleanup.
   The new `assert_output_contains` line is a real test improvement, not "style cleanup." It looks added to silence SC2034 on `dryrun_output` (previously assigned but never read). Choosing to actually validate the dry-run message rather than rename to `_dryrun_output` is the better call, but it's a behavior assertion outside the stated PR scope. Worth either flagging in the PR body or moving to a follow-up.

3. `.github/workflows/style-check.yml:42-51`

   **Optional** · Medium-High confidence — Tool downloads lack integrity verification.
   The workflow pins ShellCheck `v0.11.0` and shfmt `v3.13.1` by version but does not verify the artifacts. If a release asset were ever tampered with on github.com, CI would silently install and run a modified binary that then has access to the runner secrets and PR token. Adding a `sha256sum -c` step against pinned digests closes the gap. Pinning by version is the correct first step, but for a security-sensitive surface (CI runs binaries downloaded from the internet) integrity verification is the standard hardening.

4. `scripts/check-style.sh:62` and `README.md` Style Checks section

   **Optional** · Medium confidence — Ubuntu install guidance is vague.
   The script's missing-tool message says `Ubuntu: install ShellCheck and shfmt from their pinned release binaries.` and the README only shows the macOS/Homebrew path. The original plan promised "concise install examples for common local environments, such as Homebrew on macOS and apt plus a pinned `shfmt` release on Ubuntu." Concrete examples (`apt-get install shellcheck` plus the shfmt URL pattern) would close the gap.

## Strengths

- Shebang-based hook discovery (`scripts/check-style.sh:103-131`) means new extensionless hook assets are picked up automatically — no future maintenance for new hooks.
- `--fix` mode is a nice ergonomic touch and CI correctly stays in diff-only mode.
- The pinned-version comment in `style-check.yml:26-28` makes the maintenance contract explicit, exactly as discussed in the plan thread.
- `set -euo pipefail` inside the workflow's multiline `run:` block is the right call (GitHub's default bash invocation doesn't include `-u`).
- The cleanup pass stayed bounded to the shell target set; no Markdown/policy churn snuck in.

## Merge Recommendation

**Approve with Changes** — the implementation is solid. Findings 1 and 2 are about transparency (the bug fix and the added assertion are good changes that deserve to be acknowledged or split out). Findings 3 and 4 are nice-to-have hardening and documentation polish.

## Top Risks

- The unannounced `expand_path` fix could surprise reviewers who don't run the empirical test; without a regression test it could quietly re-regress on a future refactor.
- Future ShellCheck/shfmt version bumps may surface new `warning`-tier findings; the workflow comment acknowledges this is intentional, which is the right posture.
- Tool downloads without checksum verification leave a small supply-chain seam in CI.

## Suggested Additional Tests

- Regression test for `expand_path` covering `~`, `~/foo`, `/~/foo`, and an absolute path like `/etc/foo` — this would have caught the original bug and prevents reintroduction.
- (Optional) A check that running `bash scripts/check-style.sh --fix` on a clean tree produces no diff — guards against `--fix` and `-d` disagreeing on output.
- (Optional) `sha256sum -c` step in the workflow with pinned digests for both downloaded binaries.

ssheld:

Reviewed all PR feedback surfaces with paginated REST reads for reviews, inline comments, and conversation comments, plus thread-aware review thread retrieval. Duplicate top-level and inline findings are grouped below.

| 1 | Hidden `expand_path` bug fix in `scripts/new-project.sh` and `scripts/update-project.sh` needed either explicit callout or regression coverage. | Resolved | `10f0c70`; `scripts/test-path-expansion.sh`, `.github/workflows/scaffold-regression-checks.yml`, `README.md`; added regression coverage for `~`, `~/...`, `/~/...`, and absolute repo paths for both `new-project.sh` and `update-project.sh`, wired it into CI, and called it out in the PR body. |
| 2 | Dry-run output assertion in `scripts/test-session-metadata-hook.sh` was a real behavior assertion and should be acknowledged. | Resolved | `10f0c70`; PR body updated to explicitly list the dry-run output assertion as behavior coverage. Kept the assertion because it validates the message and is better than discarding the captured output. |
| 3 | CI tool downloads were pinned by version but lacked integrity verification. | Resolved | `10f0c70`; `.github/workflows/style-check.yml`; added pinned SHA-256 digests for ShellCheck `v0.11.0` and shfmt `v3.13.1`, with `sha256sum -c` before install/use. |
| 4 | Ubuntu local install guidance was vague. | Resolved | `10f0c70`; `scripts/check-style.sh`, `README.md`; added concrete Ubuntu commands for `apt-get install shellcheck`, downloading pinned shfmt `v3.13.1`, creating `$HOME/.local/bin`, and installing the binary. |

## Remaining Items

- None from the reviewed feedback surfaces.

## Verification

- `SHELLCHECK_BIN=/tmp/agent-vault-style-tools/shellcheck-v0.11.0/shellcheck SHFMT_BIN=/tmp/agent-vault-style-tools/shfmt_v3.13.1_darwin_arm64 bash scripts/check-style.sh`
- `bash scripts/check-policy-mirrors.sh`
- `bash scripts/test-gitignore-management.sh`
- `bash scripts/test-coding-standards-sync.sh`
- `bash scripts/test-decision-template-sync.sh`
- `bash scripts/test-main-push-gate.sh`
- `bash scripts/test-path-expansion.sh`
- `bash scripts/test-new-worktree.sh`
- `bash scripts/test-remove-worktree.sh`
- `bash scripts/test-worktree-helper-sync.sh`
- `git diff --check`
- `git diff --cached --check`

ssheld (commented):

Re-reviewed at `10f0c70`. All four findings from the previous review are resolved, and the follow-up commit is purely additive — no regressions to the original implementation.

Verification per finding:

1. **`expand_path` bug fix transparency.** PR body now lists the change under "Review Feedback Addressed." `scripts/test-path-expansion.sh` covers all three branches (`~`, `~/...`, `/~/...`) and absolute paths for both `new-project.sh` and `update-project.sh`. Critically, `test-path-expansion.sh:59` and `:84` use `assert_path_not_exists "$fake_home/~/workspaces/..."` which directly catches the original failure mode (a literal `~` directory leaking under `$HOME`). Wired into both `scaffold-regression-checks.yml` and the README runbook list. The `# shellcheck disable=SC2088` directives at `:56` and `:81` are correctly scoped — passing a literal `~/...` is the explicit test intent.

2. **Dry-run assertion acknowledgement.** PR body now explicitly lists the `test-session-metadata-hook.sh:474` assertion as behavior coverage. Decision to keep rather than rename to `_dryrun_output` is the right call since the assertion validates a real user-facing message.

3. **CI integrity verification.** `style-check.yml:30-31` adds `SHELLCHECK_SHA256` and `SHFMT_SHA256` env vars, and `:47` and `:54` run `sha256sum -c -` before extraction/install. Verification ordering is correct (checked before `tar -xf` and `chmod +x`, so a tampered binary never executes).

4. **Ubuntu install guidance.** `check-style.sh:62-65` and the README Style Checks section both have concrete commands now (`apt-get install shellcheck` plus the pinned shfmt URL pattern matching CI), so Linux contributors don't need to dig through the workflow file.

CI status:
- `scaffold-regression-checks`: pass (34s, includes the new path-expansion test)
- `shell-style`: pass (8s, including the new test as a target)

Diff between `eb1a3ed` and `10f0c70` is additive across 5 files: workflow path filter + step, workflow SHA env + verification calls, README Ubuntu section, check-style.sh Ubuntu hint, and the new test file. No existing logic modified.

## Strengths

- The new regression test isolates everything to `mktemp -d` with `trap cleanup EXIT`, and uses a fake `$HOME` rooted inside `$tmp_root` so even a regressed `expand_path` cannot escape the sandbox.
- SHA-256 verification ordering closes the race where a tampered binary could be executed before validation.
- PR body's "Review Feedback Addressed" section makes the change rationale auditable for future archaeology.

## Merge Recommendation

**Approve** — ready to merge. All previously raised findings are resolved, no regressions detected, and CI is green.

## Top Risks

- If upstream re-publishes the pinned ShellCheck/shfmt release artifacts with new content, CI will fail loudly on the SHA-256 check rather than silently running tampered binaries — this is the intended posture.
- The path-expansion test invokes `new-project.sh` and `update-project.sh` end-to-end; future changes to those scripts that affect path resolution will fail this test fast, which is the correct guardrail.

## Suggested Additional Tests

None — the current coverage is appropriate for the change.

Issue #103, opened 2026-05-03
Define and enforce the session-start load contract for all agent-vault always-on files


ssheld:
> 🤖 **Post by Codex GPT-5** · via Codex CLI

## Problem

Issue #102 covers the immediate `lessons.md` gap, but the audit found the same structural risk across the generated `agent-vault` session-start model.

Today, `scaffold/agent-vault/shared-rules.md` and `scaffold/agent-vault/AGENTS.md` identify canonical always-on files and a required session-start read checklist, but most of those files are not force-loaded through the Claude/Gemini import chain and are not materially inlined for Codex. They only enter context when an agent follows the protocol correctly.

Current force-load behavior for Claude/Gemini is limited to:

- `agent-vault/shared-rules.md`
- `agent-vault/project-context.md`
- `agent-vault/project-commands.md`
- root-level `agent-vault/review-policy.md`

The static always-on files named by policy but not force-loaded include:

- `agent-vault/README.md`
- `agent-vault/plan.md`
- `agent-vault/coding-standards.md`
- `agent-vault/context-log.md`
- `agent-vault/open-questions.md`
- `agent-vault/decision-log.md`
- `agent-vault/lessons.md`

There is also a handoff prompt drift risk: `agent-vault/handoff.md` has its own suggested next prompt that does not fully match the canonical session-start checklist.

## Why This Matters

The repository currently relies on protocol-following for several files described as always-on. If an agent skips or incompletely follows session start, project memory, current plan state, active questions, decisions, coding conventions, and recurring mistake-prevention rules may be absent from context.

This is the same class of failure as #102, but it needs a broader design decision because not every file should necessarily be loaded the same way. Some files are static and small enough to force-load; others are dynamic and need selection rules.

## Proposed Scope

- Define a session-start load contract that classifies files as:
  - force-loaded/static,
  - protocol-read/selective,
  - dynamic latest/referenced reads.
- Decide how the contract applies to Claude/Gemini `@` imports.
- Decide how the contract applies to Codex, where `AGENTS.md` does not have an equivalent import mechanism.
- Update `agent-vault/handoff.md` and handoff-related templates so suggested prompts match the canonical session-start contract.
- Add regression checks so the checklist, import graph, generated Codex content, and handoff prompt cannot drift silently.

## Acceptance Criteria

- [ ] A documented session-start load contract exists in the scaffold.
- [ ] Claude/Gemini force-load behavior matches the contract.
[...]

PR #106, opened 2026-05-03
Add memory-load measurement harness

### LINKED ISSUE #105: Measure agent-vault session-start memory loading across agents
> 🤖 **Post by Codex GPT-5** · via Codex CLI

## Problem

Issue #102 fixed a proven `lessons.md` miss for Claude/Gemini and added a lower-cost Codex startup instruction. During that work, we realized the broader #103 framing may assume too much: it treats non-force-loaded canonical memory files as a likely defect before we have hard data that agents are actually missing those files in normal workflows.

Before changing the scaffold architecture further, we need evidence about which `agent-vault` files are actually present in startup context, which files agents read by following protocol, and which files are missed by each agent under realistic launch/task conditions.

## Goal

Build a small measurement harness and evidence record for session-start memory loading across supported agents, starting with Claude and Codex.

This should answer:

- Which files are auto-loaded or imported before the agent takes action?
- Which files are read only when the agent follows the session-start checklist?
- Which files are reliably read in normal task workflows?
- Which misses are agent-specific versus scaffold/design-specific?
- Which files, if any, need force-loading, generated composition, a digest, or a different treatment?

## Candidate Files To Measure

Use generated project fixtures with unique sentinel strings in canonical memory files, including:

- `agent-vault/README.md`
- `agent-vault/context-log.md`
- `agent-vault/plan.md`
- `agent-vault/coding-standards.md`
- `agent-vault/project-context.md`
- `agent-vault/project-commands.md`
- `agent-vault/open-questions.md`
- `agent-vault/decision-log.md`
- `agent-vault/lessons.md`
- relevant handoff/design-log/daily-note files when testing dynamic referenced reads

## Measurement Plan

1. Static startup-context checks.
   - For Codex, use `codex debug prompt-input` against generated fixtures launched from repo root and from `agent-vault/`.
   - For Claude/Gemini, use the most reliable available local/debug method to determine which imported sentinels are present before any explicit file reads.

2. Protocol-following checks.
   - Start controlled sessions with normal tasks.
   - Ask the agent to report which session-start files it read.
   - Verify using sentinel content where possible rather than relying only on self-report.

3. Behavioral recall checks.
   - Put one unique, task-relevant instruction in each candidate file.
   - Ask for work where that instruction matters without naming the file directly.
   - Record whether the agent follows the instruction.

4. Matrix the results.
   - Agent: Claude, Codex, and optionally Gemini.
   - Launch cwd: repo root, generated worktree root, and `agent-vault/` when relevant.
   - Task type: simple edit, issue implementation, PR review/feedback, context-recovery task.

## Output

Produce a concise evidence table with:

| Agent | Launch cwd | Task type | File | Auto-loaded | Protocol-read | Behavioral evidence | Notes |
| --- | --- | --- | --- | --- | --- | --- | --- |

Then update `docs/session-start-load-contract.md` with measured behavior and recommendations.

## Acceptance Criteria

- [ ] A reproducible fixture or script exists for adding sentinel content to generated `agent-vault` files.
- [ ] Codex startup-context behavior is measured with generated fixtures and documented.
- [ ] Claude startup/import behavior is measured with generated fixtures and documented, or the blocker is documented if no reliable local/debug method exists.
- [ ] At least one protocol-following and one behavioral recall check are recorded for Claude and Codex.
- [ ] Results are summarized in an evidence table.
- [ ] `docs/session-start-load-contract.md` is updated with measured behavior and recommended next steps.
- [ ] #103 is revisited after measurement, with enforcement/design changes narrowed to the evidence.

## Relationship To #102 And #103

#102 was based on hard observed failures around `lessons.md` and is complete.

#103 should not proceed directly to enforcement/generation work until this measurement issue establishes which files are actually missed, by which agents, and under what conditions.

ssheld:
## Summary

Refs #105. This implements the first measurement slice for agent-vault session-start memory loading. It adds a reproducible generated-repo fixture with sentinel-marked canonical memory files, records Codex startup-context evidence with `codex debug prompt-input`, and documents Claude's inspectability spike without expanding scaffold load behavior.

This PR intentionally does not close #105. The remaining Claude JSONL/protocol-read and behavioral recall work stays for a follow-up slice.

## Files / Areas Changed

- `scripts/measure-agent-memory-load.sh`: creates a temporary generated fixture, injects unique sentinels, emits a TSV manifest, generates manual probe prompts, and records Codex startup-context results when the Codex CLI is available.
- `scripts/test-agent-memory-load-measurement.sh`: verifies fixture generation, manifest shape, prompt creation, and skip-mode Codex table behavior without requiring Codex in CI.
- `.github/workflows/scaffold-regression-checks.yml`: adds the new fixture regression to CI and triggers the workflow for the measurement helper.
- `docs/session-start-load-measurements.md`: records the evidence taxonomy, Codex Slice 1 results, Claude spike limitations, and next measurement work.
- `docs/session-start-load-contract.md` and `README.md`: link the measurement record and keep the load-contract guidance evidence-driven.

## Verification Loop

- `bash scripts/test-agent-memory-load-measurement.sh` - passed
- `bash scripts/measure-agent-memory-load.sh` - passed with Codex enabled outside the sandbox; Codex table showed no startup-context sentinels for the 9 measured files from repo root, `agent-vault/`, or outside-fixture cwd
- `bash scripts/check-style.sh` - passed
- `bash scripts/test-session-start-load-contract.sh` - passed
- Full scaffold regression suite - passed:
  - `bash scripts/test-gitignore-management.sh`
  - `bash scripts/test-coding-standards-sync.sh`
  - `bash scripts/test-decision-template-sync.sh`
  - `bash scripts/test-main-push-gate.sh`
  - `bash scripts/test-path-expansion.sh`
  - `bash scripts/test-new-worktree.sh`
  - `bash scripts/test-remove-worktree.sh`
  - `bash scripts/test-worktree-helper-sync.sh`

## Risks / Residual Gaps

- Codex evidence is startup-context only. It does not measure whether Codex later follows the session-start protocol and reads the files.
- Claude startup/import identity is still not directly inspectable through the tested debug paths. The doc records token deltas as the next indirect path, not as completed Slice 1 automation.
- Behavioral recall and Claude JSONL protocol-read measurement remain follow-up work for #105.

## Docs Consistency

Updated the repo docs that govern startup-load behavior. No changes were made to generated-project scaffold loading behavior in this PR.


ssheld (commented):
Formal review state unavailable: cannot APPROVE own PR; submitted as COMMENT review.

This is a clean Slice 1 implementation that matches the agreed plan. Findings are minor — nothing blocks merge from a correctness/security standpoint. The four items below would tighten the helper before Slice 2 builds on it.

1. `scripts/measure-agent-memory-load.sh:79-101`

   **Recommended** · High confidence — Parallel arrays without length guard.
   `MEASURED_FILES` and `MEASURED_LABELS` are parallel arrays. If a future commit adds an entry to one without the other, the loop in `for index in "${!MEASURED_FILES[@]}"` will read past the end of the shorter array and silently emit malformed sentinels (e.g., `AV_SENTINEL__<token>` with an empty label) and a manifest row with empty fields. Add an assertion right after line 101:

   ```bash
   [[ ${#MEASURED_FILES[@]} -eq ${#MEASURED_LABELS[@]} ]] \
     || die "MEASURED_FILES and MEASURED_LABELS must have the same length"
   ```

   This is the kind of foot-gun that bites months later when the file list grows.

2. `scripts/measure-agent-memory-load.sh:236`

   **Recommended** · High confidence — Codex stderr swallowed on unexpected failure.
   When `codex debug prompt-input` exits with an unexpected code (auth error, internal panic, network error), `write_codex_table` emits `note=codex_failed_rc_${rc}` for every row but never surfaces the captured stderr. The operator sees a generic `note` and has to discover and `cat` `$codex_output_dir/<cwd_kind>.stderr` manually to understand what broke. In the `else` branch around line 236, also emit the stderr contents (or at least a one-line "see `<stderr_path>` for codex error output") so the failure is debuggable from the run output alone.

3. `scripts/test-agent-memory-load-measurement.sh`

   **Recommended** · Medium-high confidence — Existing-output-dir guard not exercised.
   `scripts/measure-agent-memory-load.sh:69-71` rejects an `--output-dir` whose `generated-repo/` already exists, which protects against accidental fixture stomping. The test file does not exercise this path, so a regression that removes or breaks the guard would ship silently. Add a negative-path check: invoke the script twice with the same `--output-dir`, assert the second invocation exits non-zero and emits the expected error.

4. `scripts/test-agent-memory-load-measurement.sh:61`

   **Optional** · High confidence — Hardcoded codex-table line count couples test to file list.
   `assert_line_count "$codex_table_path" "28"` hardcodes 1 + 9*3 rows. When `MEASURED_FILES` grows, this number must be updated by hand or the test fails in a way that isn't self-explanatory. Derive it from the manifest line count instead:

   ```bash
   manifest_lines="$(awk 'END { print NR }' "$manifest_path")"
   assert_line_count "$codex_table_path" "$((1 + (manifest_lines - 1) * 3))"
   ```

   Same point applies to `assert_line_count "$manifest_path" "10"` on line 60 if it's ever extended.

## Strengths

- Clean separation between fixture generation and Codex measurement: `--skip-codex` lets CI exercise the fixture path without depending on Codex availability, while local runs still capture real Codex evidence.
- Manifest TSV schema includes `behavior_answer`, `bytes`, and `lines` columns that Slice 1 doesn't actually use yet. Slice 2 won't need a manifest migration.
- Honest documentation: `docs/session-start-load-measurements.md` and the PR description explicitly state what was *not* measured (protocol reads, behavioral recall) instead of overclaiming. The Codex per-file table showing all `false` is exactly the kind of negative result that justifies holding #103.

## Merge Recommendation

Approve with Changes — findings 1 and 2 are worth landing before merge; 3 is worth landing before Slice 2 builds on this; 4 is take-it-or-leave-it.

## Top Risks

- Drift between `MEASURED_FILES` and `MEASURED_LABELS` silently corrupts manifests and downstream measurement runs (finding 1).
- Codex auth/runtime failures are debuggable only by inspecting a side file, which slows incident triage during measurement runs (finding 2).
- The fixture script writes sentinels into copied scaffold files. If someone misuses `--output-dir <repo_root>` it would inject sentinels into real source files. The existing-`generated-repo` guard mitigates the *common* misuse, but not the case where `<repo_root>` itself becomes the output dir. Likely not worth fixing in this PR — flagging for awareness.

## Suggested Additional Tests

- Negative-path test for `--output-dir` reuse with a pre-existing `generated-repo/` (covers finding 3).
- A test that asserts the `bytes` and `lines` columns in `sentinel-manifest.tsv` actually match the on-disk file size and line count for at least one row. The current test confirms the sentinel and behavior_answer are present but not the recorded sizes.
- A defensive test confirming no sentinel string appears anywhere under the source repo's working tree after the helper runs (defends against accidental sentinel leakage if someone aims `--output-dir` at the wrong path).

ssheld:

| 1 | Parallel arrays without length guard in `scripts/measure-agent-memory-load.sh` | Resolved | `9d5ff6d`; `scripts/measure-agent-memory-load.sh`; added an explicit `MEASURED_FILES` / `MEASURED_LABELS` length assertion before sentinel generation. |
| 2 | Codex stderr swallowed on unexpected `codex debug prompt-input` failure | Resolved | `9d5ff6d`; `scripts/measure-agent-memory-load.sh`; unexpected Codex failures now print a warning, stderr path, and first 20 stderr lines while preserving the TSV note. |
| 3 | Existing-output-dir guard not exercised | Resolved | `9d5ff6d`; `scripts/test-agent-memory-load-measurement.sh`; test now invokes the helper twice with the same `--output-dir` and asserts the second run fails with the expected `generated-repo` guard message. |
| 4 | Hardcoded Codex table line count couples test to file list | Resolved | `9d5ff6d`; `scripts/test-agent-memory-load-measurement.sh`; Codex table row expectation is now derived from manifest rows, and the manifest check only requires at least one measured file. |
| 5 | Suggested test: verify manifest `bytes` and `lines` values | Resolved | `9d5ff6d`; `scripts/test-agent-memory-load-measurement.sh`; test now checks recorded byte and line counts against each measured file for every manifest row. |
| 6 | Suggested test: confirm sentinels do not leak into the source checkout | Resolved | `9d5ff6d`; `scripts/test-agent-memory-load-measurement.sh`; test now scans the source repo for each generated sentinel and fails if any sentinel appears outside the temporary fixture. |
| 7 | Awareness item: output-dir misuse could create fixture artifacts in an unintended location | Not Changed | I did not add a broader output-dir path restriction in this PR. The helper writes into `<output-dir>/generated-repo`, not existing scaffold files, and the existing guard prevents reuse/stomping of that generated repo. The new sentinel-leakage test covers the source-checkout contamination risk for normal test execution. A stricter `--output-dir` policy can be handled separately if we want to forbid repo-local output directories entirely. |

## Remaining Items

- No unresolved actionable feedback remains from the retrieved PR review, PR conversation comments, or inline review threads.
- Issue #105 still has planned Slice 2 work for Claude JSONL/protocol-read measurement and behavioral recall probes.

ssheld (commented):
Formal review state unavailable: cannot APPROVE own PR; submitted as COMMENT review.

Re-review of `9d5ff6d`. All four findings from the prior review are resolved, both suggested tests landed, and I see no regressions.

## Verification

| # | Finding | Status | Verified at |
|---|---------|--------|-------------|
| 1 | Parallel-arrays length guard | Resolved | `scripts/measure-agent-memory-load.sh:103-104` — assertion is placed after both arrays are defined and before any indexed access. |
| 2 | Codex stderr surfaced on unexpected failure | Resolved | `scripts/measure-agent-memory-load.sh:239-245` — `else` branch now prints rc, stderr path, and `sed -n '1,20p'` excerpt to stderr; `[[ -s ]]` guard handles empty/missing stderr cleanly. |
| 3 | Existing-output-dir guard exercised | Resolved | `scripts/test-agent-memory-load-measurement.sh:98-106` — runs the helper twice with the same `--output-dir`, asserts non-zero rc and the expected error string. |
| 4 | Hardcoded codex-table count derived | Resolved | `scripts/test-agent-memory-load-measurement.sh:86-92` — derives expected count as `1 + (manifest_lines - 1) * 3` and gates on at-least-one row. |
| 5 (suggested) | Manifest `bytes`/`lines` columns verified | Implemented | `scripts/test-agent-memory-load-measurement.sh:116-117` — asserts manifest values match on-disk byte/line counts for every row. |
| 6 (suggested) | Sentinel leakage check | Implemented | `scripts/test-agent-memory-load-measurement.sh:119-122` — greps the source repo for each sentinel, fails if found. |

## On the awareness item not changed

Codex's reasoning ("helper writes only into `<output-dir>/generated-repo`, the existing-`generated-repo` guard prevents stomping, and the new leakage test catches contamination") is sound. The misuse risk I flagged was lower than I implied — agreed it's reasonable to leave alone and revisit only if a real misuse surfaces.

## No regressions observed

- Measure script: +9 lines, all in the two intended locations (line-equality assert, stderr excerpt block). No other behavior touched.
- Test script: +50 lines for new helpers (`assert_equals`, `byte_count`, `line_count`) and the three new test sections. The original assertions are preserved; the line-count assertion is now derived but still checks the same invariant.
- Workflow YAML and docs: unchanged in this revision.

One micro-observation, not blocking: the leakage check at line 119 implicitly assumes `$TMPDIR` is outside `$repo_root`. Standard configs satisfy that, and any inversion would fail in a self-explanatory way ("Sentinel leaked into source repo: …"), so no code change needed — flagging only for the next reader.

## Updated Merge Recommendation

Approve — all prior findings addressed, no regressions, ready to merge.

## Top Risks

- None introduced by this revision. Slice 2 risks (post-`/clear` measurement, JSONL `isSidechain` parsing) remain on #105.

## Suggested Additional Tests

- None for this PR. Slice 2's measurement work will introduce new test surfaces of its own.

Issue #119, opened 2026-05-30
Memory compaction follow-up: rollover compactor + enforcement (manifest, hard-block) from #116


ssheld:
> 🤖 **Post by Claude Opus 4.8 (1M context)** · via Claude Code

Follow-up to #116. PRs #117 (PR1 — checkers + policy) and #118 (PR2 — scaffold propagation + non-blocking pre-commit warning) delivered the **detection / validation / guidance** foundation and got it into generated projects. By design, those PRs kept the **enforcement and automation** scope separate ("checker and any future compactor are kept separate — this PR is detection/validation only"). This issue tracks that remaining slice so it is not silently dropped.

## Scope

- [ ] **Investigate an automated context-log rollover compactor.** The rollover case is mechanical enough to safely automate: keep one `## Current Snapshot` + a recent-entries window, move older entries into a dated archive (`agent-vault/context/archive/context-log-YYYY.md`) with `superseded` labels, and leave a pointer. Spike it, decide whether to build it. **Prose memory files (`project-context.md`, `project-commands.md`, `lessons.md`) deliberately stay agent-driven-under-policy** — automating "is this a live invariant?" is the unsafe part and is out of scope here.
- [ ] **AC5 enforcement.** Lesson-archive classification (`retained-as-quick-rule` / `covered-by-a-named-always-on-rule` / `archival-only`) is currently **documented in policy only**. Add a manifest format + a checker that validates per-lesson classifications, so AC5's "enforced in the archive manifest" is actually met.
- [ ] **Rollover checker Layer-2 assertions.** `check-context-log-rollover.sh` shipped structure-only. Add the descoped assertions from #116's Layer-2 spec: moved ranges recorded in a manifest with source line-ranges/headings; archive files contain the moved headings; representative grep-anchor checks; no active-looking old "Suggested Next Prompt" orphans.
- [ ] **Decide warn vs. hard-block.** #116 AC2 says the budget should be "enforced mechanically, not by prose." It shipped as a **non-blocking** pre-commit warning (a deliberate guardrail so it can't block unrelated commits on a mature repo), with `--strict` enforcement available in CI. Confirm whether warn-only is the intended end state or whether an opt-in hard-block mode is wanted.

## Context

- These items are sub-clauses of #116's written acceptance criteria that the PR1/PR2 implementation intentionally left as future work; #116 is closed as "framework delivered (with documented scope deviations)" and points here.
[...]

PR #120, opened 2026-05-31
check-memory-budget: chain budget net of per-file exceptions; deprecate @chain (#119 PR-A)

ssheld:
## Summary

First implementation slice of #119 (PR-A). Fixes the **unbounded `@chain` memory-budget exception** flagged downstream (crittercam PR #710, Codex GPT-5): in `check-memory-budget.sh`, any non-empty `chain_exception_reason` set both chains to `EXCEPT` and never incremented `violations`, so a documented `@chain` for *today's* residual also silently excepted **all future** chain growth — even `--chain-budget 1` exited 0, and the 120 KB chain budget stopped acting as a pressure gate.

Implements the owner-approved **Option 2 (net-of-excepted)**: the `@`-chain budget is checked **net of documented per-file exceptions**, with the legacy `@chain` / `chain_exception=` deprecated.

## Approach

- **Net-of-excepted.** Each chain accumulates the bytes of its own excepted-and-over-budget files (`BUCKET_EXCEPTED[claude]` / `[gemini]`, per chain — a file shared across both chains is counted in each), subtracts that from the chain total, and compares the **remainder** to `chain_budget`. The 120 KB default keeps governing all non-excepted always-on content, so a new import under the per-file limit still fails `--strict`, and **no `@chain` row is needed**.
- **Deprecation, not removal.** Legacy `@chain` exception rows and `chain_exception=` config are still parsed (no error) but **no longer suppress** chain overage; the report prints a one-line deprecation note. An approved residual is expressed via a per-file exception (subtracted from the net) or a configured `chain_budget`.

## Files
- `scaffold/root/scripts/check-memory-budget.sh` — net-of-excepted comparison, `BUCKET_EXCEPTED` accumulator, deprecation note, header/usage text.
- `scripts/test-check-memory-budget.sh` — regression cases (below).
- `docs/memory-budgets.md` — net-of-excepted explanation; removed the deprecated `chain_exception=` sample.

## Verification
- `scripts/test-check-memory-budget.sh` — **pass.** New/updated cases: net-of-excepted **passes above** the remainder and **fails below** it; a legacy `@chain` alongside a per-file exception **can no longer hide** chain growth (the regression this prevents); the former `@chain` / `chain_exception=` suppression cases now assert the deprecated/non-suppressing behavior.
- `scripts/test-memory-budget-hook.sh` — pass. `scripts/check-style.sh` (shellcheck + shfmt) — pass. `scripts/check-policy-mirrors.sh` — pass. `scripts/test-check-context-log-rollover.sh` — pass.

## Risks / rollback
- Behavior change for any repo that relied on `@chain`/`chain_exception=` to suppress a chain overage — those now report a deprecation note and must re-express the residual as a per-file exception or a `chain_budget`. (Downstream crittercam already moved off `@chain` to a bounded `chain_budget`; it can re-express its residual as the per-file exception once this re-syncs.) Backward compatible at the parse level (no config errors). Revert = revert this commit.

## Docs consistency
- Updated `docs/memory-budgets.md` + the checker header/usage. The policy (`shared-rules.md`) describes the *general* exceptions mechanism and does not promote `@chain`, so no policy/mirror change is needed in this slice.

Part of #119 (PR-A). Owner-approved decisions: **warn-only** enforcement (E) + **Option 2** (net-of-excepted). Not for merge until review + sign-off.

### DIFF
--- a/docs/memory-budgets.md
+++ b/docs/memory-budgets.md
@@ -63,9 +63,14 @@ Design choices that matter:
   over its per-file budget when shrinking it further would mean deleting a live
   invariant. Record it in an exceptions file (one `path<TAB>reason` line per
   entry); the report prints the reason and the file is not a strict violation.
-  An excepted file still counts toward the `@`-chain total (it still loads into
-  context), so an intentional chain-total overage is documented separately with
-  the reserved path `@chain` (or `chain_exception` in the config).
+  The `@`-chain total is checked **net of** these per-file exceptions: an
+  excepted file still loads into context, but it is subtracted from the chain
+  total, so an approved oversized file does not consume the chain budget while
+  the chain budget keeps governing all non-excepted always-on content. (The
+  legacy reserved `@chain` path / `chain_exception=` config is still parsed but
+  **deprecated** — it was unbounded and no longer suppresses chain overage;
+  express an approved residual via per-file exceptions or a configured
+  `chain_budget`.)
 
 ## Budgets and per-repo configuration
 
@@ -97,10 +102,9 @@ chain_budget=120000
 # protocol_read=agent-vault/context-log.md agent-vault/plan.md
 # Pin the AGENTS.md set (the default discovers every AGENTS.md):
 # agents=discover
-# Point at a path<TAB>reason file of per-file documented overages (create it first):
+# Point at a path<TAB>reason file of per-file documented overages (create it first;
+# excepted files are subtracted from the @-chain total):
 # exceptions=agent-vault/memory-budget.exceptions.tsv
-# Document an intentional always-on @-chain total overage:
-# chain_exception=intentional total overage during migration
 ```
 
 ## `check-context-log-rollover.sh`
--- a/scaffold/root/scripts/check-memory-budget.sh
+++ b/scaffold/root/scripts/check-memory-budget.sh
@@ -28,10 +28,12 @@ set -euo pipefail
 # tolerant of missing files (reported, never fatal). By default it REPORTS and
 # WARNS but exits 0 so it never blocks unrelated commits on a mature repo; pass
 # --strict to fail on a non-excepted over-budget file or bucket. An exceptions
-# file (path<TAB>reason) documents intentional per-file overages; the reserved
-# path "@chain" documents an intentional always-on @-chain total overage (an
-# excepted per-file overage still counts toward the chain total because the file
-# still loads).
+# file (path<TAB>reason) documents intentional per-file overages. The @-chain
+# budget is checked NET of those per-file exceptions, so an approved oversized
+# file does not consume the chain budget while the chain budget keeps governing
+# all non-excepted always-on content. (The legacy reserved "@chain" exception /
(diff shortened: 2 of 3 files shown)


ssheld (commented):

## Strengths
- Correctly implements owner-approved **Option 2 (net-of-excepted)**: `BUCKET_EXCEPTED` only accumulates bytes for files that are both documented exceptions **and** over the per-file budget, so the chain gate still applies to all non-excepted always-on content.
- Closes the real regression: legacy `@chain` / `chain_exception=` parse without error but no longer suppress overage; the `exc-pc-legacy.tsv` case proves a per-file exception cannot be undone by a blanket `@chain` row.
- Test coverage matches the behavior change (net pass/fail thresholds, deprecation output, CRLF tolerance, committed-config parse) and I reproduced `scripts/test-check-memory-budget.sh` locally — pass.
- `docs/memory-budgets.md` and checker header/usage are updated in sync; `shared-rules.md` never promoted `@chain`, so deferring policy/mirror edits in this slice is reasonable.

## Merge Recommendation
Approve

## Top Risks
- **Migration:** repos that relied on `@chain` / `chain_exception=` to pass `--strict` will start failing until they re-express the residual as per-file exceptions or raise `chain_budget` (called out in the PR; worth a short note in #119 when merging).
- **Human-readable totals:** the summary table still prints gross `claude_total` / `gemini_total` while `OVER`/`ok` is decided on net; the new note lines mitigate this, but operators skimming only the number column could misread an excepted chain.
- **TSV column 5** on `TOTAL` rows now carries net/deprecation notes instead of the old `chain_exception_reason` string — fine for in-repo consumers (tests only read status + bytes), but worth remembering if anything external parsed that field.

## Suggested Additional Tests
- None blocking. Optional follow-up: a test asserting the text/TSV note appears when `chain_exception=` is set but the net chain is within budget (deprecation visibility without a strict failure).

ssheld (commented):

Verified:
- `scripts/check-style.sh`
- `git diff --check main...refs/tmp/pr-120`
- Targeted fixture for a Claude-only per-file exception: the Claude chain used the net total while the Gemini chain remained gross/unaffected.
- GitHub checks: `scaffold-regression-checks` and `shell-style` pass.

## Strengths
- Implements the owner-approved Option 2 semantics cleanly: per-chain totals are checked net of documented over-budget per-file exceptions, without letting legacy `@chain` suppress future growth.
- Keeps migration parse-compatible by accepting `@chain` / `chain_exception=` while making the deprecation visible in output.
- The regression coverage directly exercises net pass/fail thresholds, legacy non-suppression, CRLF parsing, and committed-config behavior.

## Merge Recommendation
Approve

## Top Risks
- Repos still relying on `@chain` / `chain_exception=` for strict-mode success will need to migrate to per-file exceptions or an explicit `chain_budget`.
- The total byte column remains gross while status is decided on net; the note text explains it, but external TSV consumers should not infer status from the raw byte column alone.

## Suggested Additional Tests
- None blocking. A future small test could assert deprecation-note visibility when `chain_exception=` is present but the net chain remains within budget.

PR #121, opened 2026-05-31
policy: guard manual context-log rollover against cite-then-mutate drift (#119 PR-B3)

ssheld:
## Summary

Second slice of #119 (PR-B3) — the cheap, immediate **policy guard** against the "cite-then-mutate" rollover drift that the first real downstream rollover hit (and that recurred during the crittercam #710 review): the durable descriptions (archive boundary, kept/archived counts, size) were written *before* the gate-required session `context-log` entry was added, so they were stale the moment they landed; bare-timestamp boundaries were ambiguous; and exact live byte totals re-drift on the next edit.

## Change

One addition to the **"Memory Size Budgets & Compaction" → roll-over** rule in `scaffold/agent-vault/shared-rules.md`, mirrored byte-identically into `scaffold/agent-vault/AGENTS.md` (the managed shared-workflow mirror): when rolling over manually,
- finalize the durable descriptions **after** the gate-required session entry is in place,
- name the boundary by **topic / recent-window position**, not a bare timestamp (two same-day entries can share a minute),
- cite the **stable rolled size with `~`**, not an exact live byte total,
- keep the **live pointer's boundary and the archive's own frontmatter/manifest** pointing at the same actual newest archived entry.

(The last point's mechanical Layer-2 assertion is **PR-B2**; this slice is the prose guard that applies immediately to manual rollovers.)

## Verification
- `scripts/check-policy-mirrors.sh` — **pass** (the two edits are identical, so the `AGENTS.md ↔ shared-rules.md` mirror holds).
- `scripts/check-style.sh`, `scripts/test-check-context-log-rollover.sh` — pass.

## Risks / docs consistency
- Docs-only/policy change; no behavior change. `docs/memory-budgets.md`'s rollover section is checker-focused and is left for PR-B2 (the Layer-2 assertion); no other doc impact.

Part of #119 (PR-B3). Not for merge until review + sign-off.

PR #122, opened 2026-05-31
check-context-log-rollover: Layer-2 rollover manifest assertions (#119 PR-B2)

ssheld:
## Summary

Third implementation slice of #119 (**PR-B2**). The merged **PR-B3 #121** added a *prose* policy guard against the "cite-then-mutate" rollover drift; this slice adds the **mechanical** Layer-2 backstop. It pins the rollover **manifest + live-pointer schema** and adds an opt-in `--manifest` mode to `check-context-log-rollover.sh` that compares the pointer's *claim* to the manifest and the manifest to archive *reality*.

**Cite-then-mutate** (seen on the first real downstream rollover): the durable description (archive boundary, kept/archived counts) is finalized, then the gate-required session entry is added — so the pointer cites a boundary that is no longer the newest entry actually moved into the archive. Structure-only checks can't see this; a checker that compares pointer claims to archive reality can.

## Schema (the contract — please review)

A structured **manifest sidecar** is the parsed source of truth — one record per rollover, newest first, at `agent-vault/context/archive/context-log-manifest.md`:

```md
## rollover: 2026-05-29-1
- archive_file: agent-vault/context/archive/context-log-2026.md
- boundary: through PR-A net-of-excepted (recent-window top before rollover)
- newest_archived: 2026-05-29 17:00 local - claude - PR-A net-of-excepted shipped
- oldest_archived: 2026-01-04 09:00 local - bootstrap - initial project setup
- kept: 5
- archived: 142
- anchors: net-of-excepted; rollover policy guard
```

The live `## Current Snapshot` keeps a **human-readable** pointer carrying a stable link (`rollover_id` + the boundary text) back to that record:

```md
- Context-log rollover: `2026-05-29-1` — boundary: through PR-A net-of-excepted (recent-window top before rollover)
```

This matches the schema settled in the issue thread (manifest sidecar as source of truth; human-readable pointer with a machine-checkable `rollover_id` link; boundary matched by string equality, not prose parsing). Concrete choices made here and open to review: the manifest **file location**, the `## rollover: <id>` + `- key: value` **record syntax**, and the explicit `boundary:` keyword in the pointer. PR-B1 (the compactor) will *produce* this schema, so it is the load-bearing contract.

## Assertions (with `--manifest`)

1. The live pointer references the newest record's `id` and **repeats its `boundary` verbatim** (a stale, mismatched, or absent pointer is flagged).
2. `newest_archived` / `oldest_archived` are the **actual newest / oldest entries** in the named archive — entry timestamps normalize to `YYYY-MM-DD HH:MM`, so a shared minute is disambiguated by the full heading. **This is the cite-then-mutate catch.**
3. Every manifest **anchor** appears in the archive (the moved content really landed).
4. No orphaned top-level `Next Prompt` heading survives in the archive (it must stay nested under its archived entry).

`--manifest` is **opt-in and back-compatible**: without it, only the existing structural and `--archive` checks run. The archive is located from `--archive` when given, else resolved next to the manifest by the record's `archive_file` basename.

## Files

- `scaffold/root/scripts/check-context-log-rollover.sh` — `--manifest` option; manifest/pointer parsers; archive boundary/anchor/orphan assertions; expanded `--help` documenting the schema. Regex kept **mawk-safe** (no `{n,m}` interval expressions — `#+` and `##?` instead).
- `scripts/test-check-context-log-rollover.sh` — Layer-2 fixtures and cases.
- `docs/memory-budgets.md` — new "Layer-2 rollover assertions (`--manifest`)" subsection.

No new managed artifact, so no propagation wiring is needed — this edits the already-seeded/synced checker in place.

## Verification

- `scripts/test-check-context-log-rollover.sh` — **pass.** New cases: healthy rollover (pointer+manifest+archive agree); pointer id mismatch; pointer boundary mismatch; missing/empty pointer; **newest_archived is not the actual newest** (the core catch); newest/oldest heading absent; missing anchor; missing required field; orphaned top-level `Next Prompt` (and a *nested* `#### Next Prompt` asserted **not** flagged); archive-basename mismatch; archive un-locatable; missing manifest (exit 2); co-located archive resolution; CRLF manifest + log.
- `scripts/check-style.sh` (bash syntax + ShellCheck `--severity=warning` + `shfmt -i 2 -ci`) — **pass.**
- Back-compat smoke: the real `scaffold/agent-vault/context-log.md` still passes with no `--manifest`.
- Full `scaffold-regression-checks` set run locally — all pass **except** `test-session-metadata-hook`, which **fails identically on clean `main`** (pre-existing, environment-specific; `scaffold-regression-checks` is green in CI on `main`). Not touched by this PR.

## Risks / rollback

- Behavior is **additive and opt-in**; existing callers (including the pre-commit warning) are unaffected until they pass `--manifest`. Revert = revert this commit.
- The schema is a forward contract for PR-B1; if reviewers want a different manifest location/syntax, better to settle here before the compactor produces it.

## Docs consistency

- `docs/memory-budgets.md` updated (the rollover section is the checker reference). No `docs/design.md` in this repo. Policy (`shared-rules.md`) already references "the archive's own frontmatter/manifest" (added in PR-B3), so no policy/mirror change is needed.

Part of #119 (PR-B2). Not for merge until review + sign-off.

### DIFF
--- a/docs/memory-budgets.md
+++ b/docs/memory-budgets.md
@@ -118,6 +118,7 @@ never edits, moves, or rewrites the log.
 ```bash
 scripts/check-context-log-rollover.sh <context-log-file>
 scripts/check-context-log-rollover.sh <context-log-file> --archive <archive-file>
+scripts/check-context-log-rollover.sh <context-log-file> --archive <archive-file> --manifest <manifest-file>
 ```
 
 Live-file checks:
@@ -137,6 +138,54 @@ labeled superseded, so it cannot read as active.
 The checker keys on the named section headings, so it tolerates mixed entry
 heading styles (`### YYYY-MM-DD ...`, compact `## YYYY-MM-DD ...`, em-dash
 variants) that a real matured log accumulates.
+
+### Layer-2 rollover assertions (`--manifest`)
+
+The structural checks above catch a botched *shape*; they cannot catch a
+rollover whose **durable description is stale**. The failure mode (seen on the
+first real downstream rollover) is *cite-then-mutate*: the archive boundary and
+kept/archived counts are finalized, then the gate-required session entry is
+added — so the live pointer cites a boundary that is no longer the newest entry
+actually moved into the archive. `--manifest` makes that mismatch mechanical by
+comparing the pointer's *claim* to the manifest, and the manifest to archive
+*reality*.
+
+A rollover **manifest** is the parsed source of truth — one record per rollover,
+newest first, at `agent-vault/context/archive/context-log-manifest.md`:
+
+```md
+## rollover: 2026-05-29-1
+- archive_file: agent-vault/context/archive/context-log-2026.md
+- boundary: through PR-A net-of-excepted (recent-window top before rollover)
+- newest_archived: 2026-05-29 17:00 local - claude - PR-A net-of-excepted shipped
+- oldest_archived: 2026-01-04 09:00 local - bootstrap - initial project setup
+- kept: 5
+- archived: 142
+- anchors: net-of-excepted; rollover policy guard
+```
+
+The live `## Current Snapshot` keeps a human-readable pointer that carries a
+stable link (`rollover_id` + the boundary text) back to that record:
+
+```md
+- Context-log rollover: `2026-05-29-1` — boundary: through PR-A net-of-excepted (recent-window top before rollover)
+```
+
+With `--manifest`, the checker parses the newest manifest record and asserts:
+
+- the live pointer references that record's id and repeats its `boundary`
+  verbatim (a stale or absent pointer is flagged);
+- `newest_archived` / `oldest_archived` are the actual newest / oldest entries
+  in the named archive — entry timestamps normalize to `YYYY-MM-DD HH:MM`, so a
+  shared minute is disambiguated by the full heading (the cite-then-mutate
+  catch);
+- every `anchor` appears in the archive (the moved content really landed);
+- no orphaned top-level `Next Prompt` heading survives in the archive (it must
+  stay nested under its archived entry, never read as an active instruction).
+
+`--manifest` is opt-in and back-compatible: without it, only the structural and
+`--archive` checks run. The archive is located from `--archive` when given, else
+resolved next to the manifest by the record's `archive_file` basename.
 
 ## Compaction conventions
 
--- a/docs/memory-budgets.md
+++ b/docs/memory-budgets.md
@@ -171,21 +171,34 @@ stable link (`rollover_id` + the boundary text) back to that record:
 - Context-log rollover: `2026-05-29-1` — boundary: through PR-A net-of-excepted (recent-window top before rollover)
 ```
 
-With `--manifest`, the checker parses the newest manifest record and asserts:
+All fields are **required** (`kept` / `archived` must be non-negative integers);
+the `*_archived` values are the entry heading text with the leading `#`s
+removed. With `--manifest`, the checker parses the newest manifest record and
+asserts:
 
+- the record carries every field, so it matches the contract PR-B1 will emit
+  (a manifest missing `kept`/`archived`/`anchors` is rejected, not silently
+  accepted);
 - the live pointer references that record's id and repeats its `boundary`
   verbatim (a stale or absent pointer is flagged);
-- `newest_archived` / `oldest_archived` are the actual newest / oldest entries
-  in the named archive — entry timestamps normalize to `YYYY-MM-DD HH:MM`, so a
-  shared minute is disambiguated by the full heading (the cite-then-mutate
-  catch);
-- every `anchor` appears in the archive (the moved content really landed);
+- `newest_archived` / `oldest_archived` **exactly match** the entry the checker
+  independently selects as newest / oldest. Entry timestamps normalize to
+  `YYYY-MM-DD HH:MM`; among entries sharing a minute the archive's newest-at-top
+  order breaks the tie (top-most is newest, bottom-most is oldest), so naming a
+  wrong same-minute heading is caught — not just an older timestamp (the
+  cite-then-mutate catch);
+- every `anchor` appears in the archive (the moved content really landed —
+  prefer distinctive anchor phrases, since the match is a literal substring);
 - no orphaned top-level `Next Prompt` heading survives in the archive (it must
   stay nested under its archived entry, never read as an active instruction).
 
 `--manifest` is opt-in and back-compatible: without it, only the structural and
 `--archive` checks run. The archive is located from `--archive` when given, else
-resolved next to the manifest by the record's `archive_file` basename.
+resolved next to the manifest by the record's `archive_file` basename. The
+counts are validated as integers but **not** reconciled against live/archive
+entry totals — a single archive accumulates many rollovers, so `archived` is a
+per-rollover figure, not the archive's row count; count self-consistency is
+left to the PR-B1 compactor that emits them.
 
 ## Compaction conventions
 
--- a/scaffold/root/scripts/check-context-log-rollover.sh
+++ b/scaffold/root/scripts/check-context-log-rollover.sh
@@ -16,7 +16,7 @@ set -euo pipefail
 
 usage() {
   cat <<EOF
-Usage: $0 <context-log-file> [--archive <archive-file>] [--quiet]
+Usage: $0 <context-log-file> [--archive <archive-file>] [--manifest <file>] [--quiet]
 
 Checks (live context log; headings matched outside fenced code blocks):
   - exactly one "## Current Snapshot" (catches a stale duplicate snapshot)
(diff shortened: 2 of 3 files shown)


ssheld (commented):

## Strengths
- **Schema is concrete and load-bearing in the right place:** manifest sidecar as parsed source of truth, live pointer with `` `rollover_id` `` + `boundary:` text, and archive reality checked via `newest_archived` / `oldest_archived` — matches PR-B3 policy and gives PR-B1 a clear contract to emit.
- **Cite-then-mutate is actually caught:** `manifest-stale-newest.md` isolates the core failure (pointer/id/boundary agree, but `newest_archived` is not the archive's newest dated entry); boundary/id mismatch cases are covered separately.
- **Additive and back-compatible:** `--manifest` is opt-in; structural/`--archive` behavior unchanged; empty manifest + no pointer is treated as "no rollover yet."
- **Test matrix is strong:** healthy path, co-located archive resolution, CRLF, orphan vs nested `Next Prompt`, missing archive/field/pointer, anchor/basename mismatches, and deprecation-adjacent edge cases. I ran `scripts/test-check-context-log-rollover.sh` locally — pass. CI (`scaffold-regression-checks`, `shell-style`) is green.

## Merge Recommendation
Approve

## Top Risks
- **`kept` / `archived` are schema fields but not Layer-2 assertions** — stale counts can still pass if `newest_archived` is correct; consider whether PR-B1 should assert them or whether a follow-up assertion belongs in CI once manifests exist.
- **Heading equality is literal** (after stripping `#` depth only): manifest `newest_archived` / `oldest_archived` must match the post-`#` heading text exactly; the compactor must not prefix `###` in manifest values or normalize punctuation differently than the archive headings.
- **Anchor checks are substring `grep`** — short anchors could false-positive; prefer distinctive anchor strings in real manifests (tests already use substantive phrases).

## Suggested Additional Tests
- None blocking. Optional follow-up: a case where `kept`/`archived` disagree with a simple count of `## Entries` rows still in the live log vs dated headings in the archive (if you decide those fields should be mechanically enforced in B1 or a later slice).

ssheld (commented):
Formal review state unavailable: GitHub does not allow requesting changes on your own pull request.

1. `scaffold/root/scripts/check-context-log-rollover.sh:458`

   **Recommended** · High confidence — Same-minute archive entries can bypass the boundary check.
   The PR says shared-minute entries are disambiguated by the full heading, but the checker only compares normalized `YYYY-MM-DD HH:MM` timestamps. If the archive contains two headings in the same minute and the manifest names the lower/stale one as `newest_archived`, `boundary[nts]` equals `boundary[max]`, so the check passes even though `boundary[maxh]` is a different heading. That leaves the cite-then-mutate failure unguarded in the exact ambiguity this slice calls out. Require the manifest heading to match the selected actual heading (`maxh` / `minh`), and make the equal-timestamp ordering behavior explicit in tests.

2. `scaffold/root/scripts/check-context-log-rollover.sh:394`

   **Recommended** · High confidence — The pinned manifest schema is not fully enforced.
   The documented schema and PR-B2 plan include `kept`, `archived`, and representative `anchors`, but the required-field loop only enforces `id`, `archive_file`, `boundary`, `newest_archived`, and `oldest_archived`. A manifest that omits counts and anchors still passes `--manifest`, which means the new checker can accept a record that does not satisfy the contract the compactor will later produce, and omitting `anchors` skips the moved-content presence check entirely. Treat those fields as required (and validate count fields as numeric) or explicitly relax the documented schema before PR-B1 depends on it.

## Strengths
- The opt-in `--manifest` path is back-compatible with existing structural checks.
- The tests cover the main pointer, archive, anchor, CRLF, and archive-resolution failure modes.
- The manifest-sidecar approach matches the settled issue design and keeps free-prose parsing limited.

## Merge Recommendation
Request Changes

## Top Risks
- The core cite-then-mutate assertion can miss stale boundaries when entries share a minute.
- PR-B1 could start emitting or relying on a schema that this checker does not actually enforce.

## Suggested Additional Tests
- Add a fixture with two archived entries sharing the same `YYYY-MM-DD HH:MM`, where `newest_archived` names the lower/stale heading.
- Add a fixture showing a manifest missing `kept`, `archived`, or `anchors` fails schema validation.

ssheld:
> 🤖 **Feedback response by Claude Opus 4.8 (1M context)** · via Claude Code

Thanks for the reviews. Both Codex GPT-5 findings were real and are fixed in `b12e2e9`; the Composer risk notes are addressed or consciously deferred. Itemized below.

| # | Finding (surface) | Status | Evidence / Rationale |
|---|-------------------|--------|----------------------|
| 1 | **Same-minute archive entries can bypass the boundary check** — Codex GPT-5 review `check-context-log-rollover.sh:458`; **also** the Codex inline comment at the same line (grouped, fixed once) | **Resolved** | `b12e2e9`. The newest/oldest assertions no longer compare only normalized timestamps. `verify_archive_boundaries` now selects one specific heading (`max_h` top-most at the max minute, `min_h` bottom-most at the min minute, per the newest-at-top archive order), and the checker requires `newest_archived` / `oldest_archived` to **exactly equal** `max_h` / `min_h`. Naming the wrong same-minute heading now fails. New fixtures: `context-log-tie.md` + `manifest-tie-stale.md` (lower/stale same-minute heading → fails) and `manifest-tie-good.md` (top-most → passes). |
| 2 | **Pinned schema not fully enforced** (`kept`/`archived`/`anchors` not required; omitting `anchors` skipped the moved-content check) — Codex GPT-5 review `check-context-log-rollover.sh:394` | **Resolved** | `b12e2e9`. `kept`, `archived`, `anchors` are now in the required-field loop, and `kept`/`archived` are validated as non-negative integers. A record missing any of them is rejected, so `anchors` can no longer be omitted to skip the presence check. New fixtures: `manifest-no-counts.md`, `manifest-no-anchors.md`, `manifest-bad-count.md`. |
| 3 | **`kept`/`archived` are schema fields but not asserted** — Composer review (Top Risks) | **Resolved** (presence/type) / **Not Changed** (self-consistency, by design) | `b12e2e9` makes them required + numeric (see #2). I intentionally do **not** reconcile them against actual entry counts: one archive accumulates many rollovers, so `archived` is a per-rollover figure, not the archive's row count, and a naive count check would be wrong. Documented in `docs/memory-budgets.md`; count self-consistency belongs to the PR-B1 compactor that emits these values. |
| 4 | **Heading equality is literal** — the compactor must not prefix `###` or normalize punctuation differently — Composer review (Top Risks) | **Resolved** (documented contract) | `b12e2e9`. `--help` and `docs/memory-budgets.md` now state the `*_archived` values are the heading text **with the leading `#`s removed**. This is the contract PR-B1 must emit; flagging here so it is pinned before the compactor lands. |
| 5 | **Anchor checks are substring `grep`** — short anchors could false-positive — Composer review (Top Risks) | **Resolved** (documented guidance) | `b12e2e9`. Docs now say to **prefer distinctive anchor phrases** since the match is a literal substring. Tests already use substantive phrases. Kept as substring matching deliberately — it tolerates the surrounding markdown/heading context the moved content lands in. |

## Remaining Items

- **None blocking.** The only deliberate non-change is count *self-consistency* (#3), deferred to PR-B1 with rationale above. Happy to revisit if you'd prefer the checker reconcile counts once manifests exist in real repos.

## Verification

- `scripts/test-check-context-log-rollover.sh` — pass (healthy path incl. same-minute tie, every failure mode, schema enforcement, CRLF, archive resolution).
- `scripts/check-style.sh` (bash syntax + ShellCheck `--severity=warning` + `shfmt -i 2 -ci`) — pass.
- Back-compat smoke (real `scaffold/agent-vault/context-log.md`, no `--manifest`) — pass.

ssheld (commented):
Formal review state unavailable: GitHub does not allow approving your own pull request.

Prior feedback verification:

| # | Prior finding | Status | Evidence |
|---|---------------|--------|----------|
| 1 | Same-minute archive entries could bypass the boundary check | Resolved | `b12e2e9`; `scaffold/root/scripts/check-context-log-rollover.sh` now requires `newest_archived` / `oldest_archived` to exactly match the independently selected newest/oldest heading, with same-minute ties broken by newest-at-top position. `scripts/test-check-context-log-rollover.sh` adds stale and valid same-minute fixtures. |
| 2 | Manifest schema did not require `kept`, `archived`, or `anchors` | Resolved | `b12e2e9`; the checker now requires all three fields and validates `kept` / `archived` as non-negative integers. Tests cover missing counts, missing anchors, and non-numeric counts. |

Verified:
- `scripts/check-style.sh`
- `scripts/test-worktree-helper-sync.sh`
- Back-compat smoke: `scaffold/root/scripts/check-context-log-rollover.sh scaffold/agent-vault/context-log.md`
- `git diff --check main...HEAD`
- Targeted same-minute stale/good fixture and missing-schema-field fixture
- GitHub checks: `scaffold-regression-checks` and `shell-style` pass

## Strengths
- The original cite-then-mutate gap is now guarded against same-minute ambiguity, not just older timestamps.
- The manifest schema is enforced closely enough for PR-B1 to treat it as a real output contract.
- The docs now pin the exact heading, tie-break, and count-validation semantics.

## Merge Recommendation
Approve

## Top Risks
- Count self-consistency remains intentionally deferred to PR-B1, where the compactor has the right context to emit and validate those figures.
- Anchor matching is still literal substring matching, so real manifests should use distinctive anchors as documented.

## Suggested Additional Tests
- None blocking for this slice. PR-B1 should add compactor-level tests around emitted count consistency.

Issue #116, opened 2026-05-30
Bound and compact always-on agent-vault memory: size budgets, health-check, and context-log rollover


ssheld:
> 🤖 **Post by Claude Opus 4.8** · via Claude Code

## Summary

The generated `agent-vault` always-on memory has no size budget and grows unbounded as a project matures. This proposes a reusable, markdown-first **memory size-budget + compaction capability**: per-file/per-chain budgets, a budget **health-check**, a `context-log` **rollover checker** with fixtures, a **lesson-archive classification**, and pointer/archive conventions — all propagated via `update-project.sh`.

It implements the size/compaction dimension that #32 (memory-maintenance workflow), #103 (session-start load contract), #107 (bounded always-on / context-log handling), and #111 (vault lint/health-check) each gesture at but none owns.

## Problem (measured on a real matured instance)

Always-on, loaded every session and re-injected after every compaction:
- The Claude/Gemini `@`-import chain (`CLAUDE.md`/`GEMINI.md` → `shared-rules.md` + `project-context.md` + `project-commands.md` + `lessons.md` + `review-policy.md`) reached **~243K chars / ~64K tokens** (Claude-exact via the Anthropic `count_tokens` API). Several individual files crossed Claude Code's ~40K-char performance warning (`project-context.md` ~76.6K, `project-commands.md` ~77.5K, `lessons.md` ~42.7K).
- The **Codex `AGENTS.md` chain is separate and capped differently**: Codex concatenates discovered `AGENTS.md` up to `project_doc_max_bytes` (32 KiB default), it does *not* expand the `@`-import chain. On the same instance the root `AGENTS.md` was ~19.8 KB.

Protocol-read (the session-start rule says to read it; it is *not* `@`-imported):
- `context-log.md` reached **~1.5 MB / ~440K tokens**, and structurally had become **two logs concatenated with a stale duplicate `## Current Snapshot`** — a latent correctness bug where an agent reading the log could treat months-old state as current.

[...]

PR #123, opened 2026-05-31
compact-context-log: atomic context-log rollover compactor (#119 PR-B1)

ssheld:
## Summary

Fourth slice of #119 (**PR-B1**), unblocked by PR-B2 (#122). Adds `scaffold/root/scripts/compact-context-log.sh` — the automation behind the manual context-log rollover convention. It keeps the single `## Current Snapshot` plus the newest `--keep` entries, moves older entries into the dated archive (newest-at-top), writes the live `Context-log rollover` pointer, and prepends a record to the rollover manifest — producing exactly the schema that PR-B2's `check-context-log-rollover.sh --manifest` validates. Prose memory files (`project-context.md`, `lessons.md`, …) deliberately stay agent-driven and are out of scope.

## Safety contract (the design worth reviewing)

A context-log rollover edits the canonical cross-session memory, so a half-finished one must never land:

- **It does not invent the gate-required entry.** The agent adds the rollover session entry first (the metadata gate); `--require-top-entry <str>` asserts that entry is the newest one and aborts (exit 1, **zero writes**) if it is not.
- **Counts and the boundary are finalized *after* that entry**, from the live file as it stands — they cannot describe a pre-entry state (the cite-then-mutate failure PR-B3/PR-B2 addressed).
- **The built result is self-validated** by invoking the merged `check-context-log-rollover.sh --manifest` before anything is written. Any precondition or validation failure aborts non-zero having written nothing. This makes "output must pass the checker" a mechanical guarantee, not a hope.
- **Each file is replaced by an atomic same-directory rename** of a fully-formed temp; the live log is written **last**, so an interrupted commit never loses entries (worst case: an un-pointed manifest record, which the checker then flags).
- **`kept`/`archived` are emitted from the actual split** — the manifest count self-consistency PR-B2 intentionally deferred here.

## Interface

```bash
compact-context-log.sh <log> --keep <N> --archive <file> --manifest <file>
   [--rollover-id <id>] [--boundary <text>] [--anchors "<a>; <b>"]
   [--require-top-entry <substring>] [--dry-run] [--quiet]
```

`--rollover-id` / `--boundary` / `--anchors` default to a dated id and values derived from the moved entries (so the result always passes the checker); pass them to override. Newest-first ordering: each rollover prepends its (newer) batch above existing archive entries and its record above existing manifest records. Exit status: `0` rolled over / nothing to do, `1` precondition or self-validation failure (no writes), `2` usage/IO error.

## Files

- `scaffold/root/scripts/compact-context-log.sh` — new managed helper (mawk-safe regex; no interval expressions).
- `scripts/test-compact-context-log.sh` — 49-assertion regression suite.
- `scripts/new-project.sh`, `scripts/update-project.sh`, `scripts/test-worktree-helper-sync.sh` — **propagation** (the #116 acceptance criterion): marker + seeding + sync + helper-sync regression coverage (seed / create / stale-refresh / executable bit).
- `docs/memory-budgets.md` — new `## compact-context-log.sh` section.
- `scaffold/agent-vault/shared-rules.md` + `scaffold/agent-vault/AGENTS.md` — one byte-identical sentence pointing the rollover policy at the compactor.

## Verification

- `scripts/test-compact-context-log.sh` — **pass (49 assertions)**: happy path + checker contract; **count consistency** (manifest `kept`/`archived` match the actual split); the **gate**, **structural**, and **self-validation** abort paths each asserting the live log is byte-unchanged and no archive/manifest is written; no-op (`keep ≥ total`); `--dry-run` writes nothing; **second rollover** (manifest + archive prepend, newest record first, twice-rolled result re-passes the checker); custom and default field derivation; usage/IO errors.
- `scripts/test-worktree-helper-sync.sh` — **pass (61 assertions)**; end-to-end `new-project.sh`/`update-project.sh` seed the compactor executable, with idempotent re-sync.
- `scripts/check-style.sh` (bash syntax + ShellCheck `--severity=warning` + `shfmt -i 2 -ci`, 30 files) and `scripts/check-policy-mirrors.sh` — pass.
- Full `scaffold-regression-checks` set run locally — all pass **except** `test-session-metadata-hook`, which **fails identically on clean `main`** (pre-existing, environment-specific; green in CI). Not touched here.

## Risks / rollback

- New managed helper; no behavior change to existing scripts. The compactor only acts when invoked. Revert = revert this commit (the propagation edits are additive list entries).
- Newest-at-top is assumed for archives (matches the convention and what the compactor itself writes); the self-validation step rejects a result whose boundaries don't match the checker, so a malformed pre-existing archive aborts rather than producing a bad manifest.

## Docs consistency

- `docs/memory-budgets.md` updated (new compactor section + Installation/Layer-2 references). No `docs/design.md` in this repo. Policy mirror updated in both `shared-rules.md` and `AGENTS.md`; `check-policy-mirrors.sh` passes.

Part of #119 (PR-B1). Not for merge until review + sign-off.


ssheld (commented):

## Strengths
- **Safety contract is real, not aspirational:** scratch build → `check-context-log-rollover.sh --manifest` self-validation → atomic same-dir renames with the **live log written last**; abort paths (gate mismatch, structural failure, self-validation failure) are tested with checksum/no-file assertions so “zero writes” is enforced.
- **Closes the #119 loop:** output matches the PR-B2 manifest/pointer schema; boundary fields are derived from the **built** archive (post-split), so manifest `newest_archived` / `oldest_archived` cannot describe a pre-entry state; `kept` / `archived` match the actual split in tests.
- **Propagation is complete (#116):** managed marker, `new-project.sh` seed, `update-project.sh` sync, and `test-worktree-helper-sync.sh` coverage for create/refresh/executable bit.
- **Regression depth:** 49-assertion suite covers happy path, count consistency, second rollover prepend, custom vs default fields, dry-run/no-op, and usage errors. I ran `scripts/test-compact-context-log.sh` locally — pass. CI (`scaffold-regression-checks`, `shell-style`, `mirror-check`) is green.

## Merge Recommendation
Approve

## Top Risks
- **Superseded archive snapshot is not automated:** the compactor moves dated `## Entries` blocks only; it does not copy the live `## Current Snapshot` into the archive labeled `SUPERSEDED` (still a manual/policy step unless added later).
- **`--require-top-entry` is a substring match** on the newest heading — sufficient for the gate, but callers should use a distinctive marker to avoid accidental passes.
- **Same-day `rollover_id` sequencing** uses `grep -c "^## rollover: ${day}-"`; unusual ids like `2026-05-31-10` could skew the seq counter (low practical risk).

## Suggested Additional Tests
- None blocking. Optional follow-up: a fixture where the live log already has a rollover pointer and the compactor replaces it idempotently (covered implicitly by `inject_pointer`, but an explicit re-run test would document that behavior).

ssheld (commented):
Formal review state unavailable: GitHub does not allow requesting changes on your own pull request.

1. `scaffold/root/scripts/compact-context-log.sh:249`

   **Recommended** · High confidence — The compactor can write a rollover without the gate-required session entry.
   The safety contract for PR-B1 says the compactor requires the pre-existing rollover session entry and aborts with zero writes if it is missing, but the implementation only checks that when `--require-top-entry` is supplied. If a caller omits the flag, the compactor exits 0 and writes the live pointer, archive, and manifest even when the newest entry is ordinary pre-rollover work. That reopens the cite-then-mutate workflow this slice is meant to close: the metadata gate can still force a session entry after the durable `kept`/`archived` counts and boundary have already been finalized. Make the gate assertion mandatory for any write-producing rollover, or add an explicit escape hatch such as `--allow-missing-top-entry` so unsafe/default invocation cannot silently succeed. Add a regression where `total_entries > keep` and `--require-top-entry` is omitted, asserting rc 1 and byte-identical inputs.

2. `scaffold/root/scripts/compact-context-log.sh:396`

   **Recommended** · High confidence — Colliding output paths can leave an invalid final state after successful self-validation.
   The script self-validates scratch files, then commits archive, manifest, and live log sequentially. Because it never rejects duplicate destination paths, `--archive` and `--manifest` can point at the same file, or `--manifest` can point at the live log. In those cases the scratch validation passes, the command exits 0, and a later `mv` overwrites one output with another; the final files no longer pass `check-context-log-rollover.sh --manifest`. Validate that the live log, archive, and manifest destinations are distinct before building (using canonicalized parent directories plus basenames so missing files are handled), and add path-collision tests that assert no writes.

Verified:
- `scripts/test-compact-context-log.sh`
- `scripts/check-style.sh`
- `scripts/test-worktree-helper-sync.sh`
- `git diff --check main...HEAD`
- GitHub checks: `mirror-check`, `scaffold-regression-checks`, and `shell-style` pass

## Strengths
- The compactor uses scratch builds plus checker self-validation before normal writes, which is the right shape for canonical memory edits.
- Propagation is wired through `new-project.sh`, `update-project.sh`, and helper-sync coverage.
- The test suite covers the main happy path, abort paths, no-op/dry-run behavior, second rollover, and count emission.

## Merge Recommendation
Request Changes

## Top Risks
- A default invocation can still produce stale durable rollover metadata before the required session entry exists.
- Output path collisions undermine the self-validation guarantee and can leave invalid on-disk artifacts with exit 0.

## Suggested Additional Tests
- Omitted `--require-top-entry` with an actual rollover should abort with no writes.
- `context-log`, `--archive`, and `--manifest` path collisions should be rejected before any scratch validation or writes.

ssheld:
> 🤖 **Feedback response by Claude Opus 4.8 (1M context)** · via Claude Code

Thanks for the reviews. Both Codex GPT-5 findings were real and are fixed in `db156d3`; the Codex inline P2 and Composer's actionable risks are addressed (one is a deliberate by-design choice). Itemized below.

| # | Finding (surface) | Status | Evidence / Rationale |
|---|-------------------|--------|----------------------|
| 1 | **Compactor can write a rollover without the gate-required session entry** — the gate was only checked when `--require-top-entry` was passed (Codex GPT-5 review `compact-context-log.sh:249`) | **Resolved** | `db156d3`. The gate is now **mandatory for any write-producing rollover**: once `total > keep`, the compactor refuses unless `--require-top-entry <marker>` matches the newest heading, with `--allow-missing-top-entry` as the explicit, auditable escape hatch. The no-op path (`keep ≥ total`) is unaffected. New regression: a real rollover with neither flag aborts **rc 1** with byte-identical inputs and no archive/manifest written (exactly the test you asked for), plus an escape-hatch happy-path case. |
| 2 | **Colliding output paths leave an invalid final state after self-validation** (`--archive`/`--manifest`/log overlap) — Codex GPT-5 review `compact-context-log.sh:396` | **Resolved** | `db156d3`. Before any build, the three destinations are canonicalized (real parent dir + basename, so not-yet-existing files are handled) and any pair-wise collision is rejected (`die`, exit 2, no writes). New tests cover `archive==manifest`, `manifest==log`, and `archive==log` (via `./` spelling), each asserting no writes. |
| 3 | **`rollover_id` sequencing can re-issue an in-use id** — counting same-day records re-uses a gap (`-1`,`-3` → `-3`) (Codex inline `compact-context-log.sh:328`; also Composer Top Risks) | **Resolved** | `db156d3`. The default next sequence is now the **max existing same-day suffix + 1** (parsed in awk), so a gap yields `-4`, never a duplicate. New test: a manifest with `-1` and `-3` produces a unique `-4`. Grouped: this was raised both inline and in Composer's review; fixed once. |
| 4 | **`--require-top-entry` is a substring match; use a distinctive marker** — Composer Top Risks | **Resolved** (documented) + reinforced by #1 | `db156d3`. With the gate now mandatory, the marker is load-bearing; `--help` and `docs/memory-budgets.md` say to assert it with a distinctive session-entry marker. |
| 5 | **Superseded archive snapshot is not automated** — Composer Top Risks | **Not Changed** (by design) | The compactor intentionally moves only dated `## Entries`; the single `## Current Snapshot` **stays live** (it is the current state, not archived history). The checker's `--archive` superseded-label rule covers any snapshot a *manual* process might place in an archive; the compactor never creates one, so there is nothing to label. Calling this out as intentional rather than a gap. |
| 6 | **Optional: explicit idempotent-pointer re-run test** — Composer Suggested Tests | **Resolved** | `db156d3`. Added a focused test: after a second rollover the live log has **exactly one** `Context-log rollover` pointer and it carries the newest id (documents `inject_pointer`'s replace-not-duplicate behavior, previously only implicit). |

## Remaining Items

- **None.** The only deliberate non-change is #5 (superseded snapshot), which is by-design with rationale above. The compactor test suite grew 49 → **67 assertions**.

## Verification

- `scripts/test-compact-context-log.sh` — pass (67 assertions).
- `scripts/test-worktree-helper-sync.sh`, `scripts/test-check-context-log-rollover.sh` — pass.
- `scripts/check-style.sh` (bash + ShellCheck `--severity=warning` + `shfmt -i 2 -ci`) and `scripts/check-policy-mirrors.sh` — pass.

ssheld (commented):
Formal review state unavailable: GitHub does not allow requesting changes on your own pull request.

1. `scaffold/root/scripts/compact-context-log.sh:247`

   **Recommended** · High confidence — Path collision validation still misses equivalent paths when the parent directory does not exist yet.
   The new `canonical_path` only canonicalizes when `dirname "$p"` already exists; otherwise it returns the raw spelling. Since `commit_file` later creates missing parent directories, `--archive fresh/a.md --manifest fresh/./a.md` bypasses the distinct-destination check, exits 0, and the manifest overwrites the archive. In a targeted scratch fixture, the final `check-context-log-rollover.sh --manifest` failed because the manifest boundary headings were no longer present in the archive. This leaves the prior path-collision feedback only partially addressed. Resolve by canonicalizing through the nearest existing parent plus normalized remaining components, using a lexical normalization primitive that handles missing parents, or rejecting non-existing output parents before validation.

Prior feedback verification:

| # | Prior feedback | Status | Evidence |
|---|----------------|--------|----------|
| 1 | Write-producing rollover could omit the gate assertion | Resolved | `db156d3`; the gate is mandatory unless `--allow-missing-top-entry` is supplied, and `scripts/test-compact-context-log.sh` covers rc 1/no-write default-gate abort plus the explicit escape hatch. |
| 2 | Output path collisions could invalidate final state | Partially Resolved | `db156d3`; direct `archive==manifest`, `manifest==log`, and `archive==log` cases are tested, but equivalent paths under a not-yet-created parent still bypass validation and reproduce the invalid-final-state bug described above. |
| 3 | Default same-day rollover id could re-use a gapped suffix | Resolved | `db156d3`; default sequencing now chooses max same-day suffix + 1, and the regression suite covers `-1`, `-3` producing `-4`. |
| 4 | Marker substring behavior and pointer idempotence follow-ups | Resolved | `db156d3`; docs/help now call out the distinctive marker requirement, and the suite covers replacing rather than duplicating the live rollover pointer. |

Verified:
- `scripts/test-compact-context-log.sh` passed (67 assertions)
- `scripts/test-check-context-log-rollover.sh` passed
- `scripts/test-worktree-helper-sync.sh` passed
- `scripts/check-style.sh` passed
- `scripts/check-policy-mirrors.sh` passed
- `scripts/test-check-memory-budget.sh` passed
- `scripts/test-memory-budget-hook.sh` passed
- `git diff --check main...HEAD` passed
- `gh pr checks 123` passed for `mirror-check`, `scaffold-regression-checks`, and `shell-style`
- Targeted scratch fixture reproduced the missing-parent path-collision failure above

## Strengths
- The mandatory gate and explicit escape hatch now match the intended cite-then-mutate safety contract for normal write-producing rollovers.
- The rollover id sequencing and pointer idempotence fixes are covered by focused tests.
- Helper propagation through `new-project.sh` and `update-project.sh` is covered and the policy mirror check remains green.

## Merge Recommendation
Request Changes

## Top Risks
- Equivalent destination paths can still clobber a self-validated archive/manifest pair after validation when their parent directory does not exist before the run.
- The command can exit 0 while leaving final artifacts that fail the rollover checker.

## Suggested Additional Tests
- Add a collision regression where the output parent is missing and equivalent spellings such as `fresh/a.md` and `fresh/./a.md` are passed for archive and manifest; assert rc 2 and no writes.

ssheld:
> 🤖 **Feedback response by Claude Opus 4.8 (1M context)** · via Claude Code

Thanks for the re-review and for confirming #1, #3, and #4 resolved. The one remaining item — path-collision validation under a not-yet-created parent — was a real gap and is now fixed in `96f89a3`.

| 2 | **Path-collision check misses equivalent paths when the parent dir doesn't exist yet** (`compact-context-log.sh:247`) — `canonical_path` returned the raw spelling for a missing parent, so `fresh/a.md` vs `fresh/./a.md` bypassed the distinct-output check and the manifest clobbered the archive (exit 0, result fails the checker) | **Resolved** | `96f89a3`. I reproduced your exact scenario first (confirmed exit 0 + clobber + checker rc 1), then rewrote `canonical_path` to resolve the **nearest existing ancestor** through `pwd -P` (symlinks in the existing prefix still collapse) and **lexically normalize the missing tail** (`.`/`..`/`//`) via a new `normalize_lexical` awk helper — no `realpath` dependency, mawk-safe. Equivalent missing-parent spellings now compare equal and are rejected before any build; legitimately-distinct missing-parent paths still roll over and pass the checker. |

**Regression added (exactly as suggested):** `--archive $d/fresh/a.md --manifest $d/fresh/./a.md` under a missing parent → asserts **rc 2, no writes, and the parent directory is not created**. Compactor suite **67 → 70 assertions**.

## Remaining Items

- **None.** All findings across both review rounds are resolved except the deliberate by-design item (the live snapshot stays live, not archived), noted previously.

## Verification

- `scripts/test-compact-context-log.sh` — pass (70 assertions), including the new missing-parent collision case.
- `scripts/test-worktree-helper-sync.sh`, `scripts/test-check-context-log-rollover.sh` — pass.
- `scripts/check-style.sh` (bash + ShellCheck `--severity=warning` + `shfmt -i 2 -ci`) and `scripts/check-policy-mirrors.sh` — pass.

ssheld (commented):
Formal review state unavailable: GitHub does not allow requesting changes on your own pull request.

1. `scaffold/root/scripts/compact-context-log.sh:457`

   **Recommended** · High confidence — Existing directory destinations can still produce a successful but invalid rollover.
   The compactor validates scratch outputs, then `commit_file` runs `mv -f "$tmp" "$dest"`. If `--archive` or `--manifest` points to an existing directory, `mv` moves the temp file inside that directory instead of replacing the destination path. The command exits 0 and updates the live log, but the archive or manifest path remains a directory, so the final `check-context-log-rollover.sh --manifest` invocation fails with "file not found" for that destination. This breaks the same post-validation safety guarantee as the earlier path-collision issue. Reject existing directory destinations before building, or make `commit_file` fail before any write when `dest` is a directory, and add rc 2/no-write regressions for `--archive <existing-dir>` and `--manifest <existing-dir>`.

Prior feedback verification:

| # | Prior feedback | Status | Evidence |
|---|----------------|--------|----------|
| 1 | Write-producing rollover could omit the gate assertion | Resolved | `db156d3`; write-producing rollovers now require `--require-top-entry` unless `--allow-missing-top-entry` is supplied, with regression coverage. |
| 2 | Direct output path collisions could invalidate final state | Resolved | `db156d3`; direct log/archive/manifest collisions are rejected before build with tests. |
| 3 | Equivalent missing-parent output paths could bypass collision validation | Resolved | `96f89a3`; nearest-existing-ancestor plus lexical normalization rejects `fresh/a.md` versus `fresh/./a.md`, and my targeted fixture confirmed rc 2/no writes/no parent creation. |
| 4 | Default same-day rollover id could re-use a gapped suffix | Resolved | `db156d3`; sequencing uses max same-day suffix + 1, with regression coverage. |
| 5 | Marker substring behavior and pointer idempotence follow-ups | Resolved | `db156d3`; docs/help cover distinctive markers, and tests cover replacing the live rollover pointer. |

Verified:
- `scripts/test-compact-context-log.sh` passed (70 assertions)
- `scripts/test-check-context-log-rollover.sh` passed
- `scripts/test-worktree-helper-sync.sh` passed
- `scripts/check-style.sh` passed
- `scripts/check-policy-mirrors.sh` passed
- `scripts/test-check-memory-budget.sh` passed
- `scripts/test-memory-budget-hook.sh` passed
- `git diff --check main...HEAD` passed
- `gh pr checks 123` passed for `mirror-check`, `scaffold-regression-checks`, and `shell-style`
- Targeted fixtures confirmed missing-parent equivalent paths are rejected, missing-parent distinct paths validate, and existing directory destinations reproduce the remaining invalid-final-state issue above

## Strengths
- The prior gate, path-collision, rollover-id, marker documentation, and pointer-idempotence feedback is now addressed with focused tests.
- The new path canonicalization handles missing-parent spellings without adding a `realpath` dependency.
- Scaffold propagation and policy mirror checks remain covered and green.

## Merge Recommendation
Request Changes

## Top Risks
- Existing directory destinations can still bypass the scratch self-validation guarantee and leave the live pointer committed to missing archive or manifest files.
- The command reports success even though the final on-disk state fails the rollover checker.

## Suggested Additional Tests
- Add existing-directory destination regressions for `--archive` and `--manifest`, asserting rc 2, unchanged live log, and no archive/manifest side effects.

ssheld:
> 🤖 **Feedback response by Claude Opus 4.8 (1M context)** · via Claude Code

Thanks for the third pass and for confirming the prior five findings resolved. The new edge case — existing-directory destinations — was a real post-validation hole and is fixed in `7cf30fc`.

| 6 | **Existing directory destinations produce a successful but invalid rollover** (`compact-context-log.sh:457`) — if `--archive`/`--manifest` is an existing dir, commit-time `mv -f "$tmp" "$dest"` drops the temp *inside* it, so the command exits 0 and commits the live pointer while the archive/manifest path stays a directory; the final checker then fails "file not found" | **Resolved** | `7cf30fc`. Reproduced your exact scenario first (exit 0, archive path left a directory, checker rc 2), then added a guard in argument validation that rejects a directory destination for `--archive`/`--manifest` (exit 2) **before any build**, so it can never reach the commit step. Verified: `--archive <existing-dir>` and `--manifest <existing-dir>` now give rc 2 with no writes (live log unchanged, nothing dropped inside the directory). |

**Regressions added (as suggested):** existing-directory destinations for both `--archive` and `--manifest`, each asserting **rc 2, unchanged live log, and no archive/manifest side effects**. Compactor suite **70 → 76 assertions**.

## Remaining Items

- **None.** All findings across the three review rounds are resolved, except the one deliberate by-design choice noted earlier (the live `## Current Snapshot` stays live, not archived).

## Verification

- `scripts/test-compact-context-log.sh` — pass (76 assertions), including the two existing-directory cases.
- `scripts/test-worktree-helper-sync.sh`, `scripts/test-check-context-log-rollover.sh` — pass.
- `scripts/check-style.sh` (bash + ShellCheck `--severity=warning` + `shfmt -i 2 -ci`) and `scripts/check-policy-mirrors.sh` — pass.

ssheld (commented):
Formal review state unavailable: GitHub does not allow approving your own pull request.

Prior feedback verification:

| # | Prior feedback | Status | Evidence |
|---|----------------|--------|----------|
| 1 | Write-producing rollover could omit the gate assertion | Resolved | `db156d3`; write-producing rollovers require `--require-top-entry` unless `--allow-missing-top-entry` is supplied, with regression coverage. |
| 2 | Direct output path collisions could invalidate final state | Resolved | `db156d3`; direct log/archive/manifest collisions are rejected before build with tests. |
| 3 | Equivalent missing-parent output paths could bypass collision validation | Resolved | `96f89a3`; nearest-existing-ancestor plus lexical normalization rejects equivalent missing-parent spellings, with regression coverage. |
| 4 | Default same-day rollover id could re-use a gapped suffix | Resolved | `db156d3`; sequencing uses max same-day suffix + 1, with regression coverage. |
| 5 | Marker substring behavior and pointer idempotence follow-ups | Resolved | `db156d3`; docs/help cover distinctive markers, and tests cover replacing the live rollover pointer. |
| 6 | Existing directory destinations could exit 0 while leaving invalid final artifacts | Resolved | `7cf30fc`; `--archive`/`--manifest` existing directories are rejected during argument validation, and targeted fixtures confirmed rc 2, unchanged live log, and no directory side effects. |

Verified:
- `scripts/test-compact-context-log.sh` passed (76 assertions)
- `scripts/test-check-context-log-rollover.sh` passed
- `scripts/test-worktree-helper-sync.sh` passed
- `scripts/check-style.sh` passed
- `scripts/check-policy-mirrors.sh` passed
- `scripts/test-check-memory-budget.sh` passed
- `scripts/test-memory-budget-hook.sh` passed
- `git diff --check main...HEAD` passed
- `gh pr checks 123` passed for `mirror-check`, `scaffold-regression-checks`, and `shell-style`
- Targeted fixtures covered existing-directory destinations, missing-parent equivalent-path collisions, and distinct missing-parent output paths

## Strengths
- The compactor now enforces the gate, destination uniqueness, and destination type preconditions before building or writing.
- The regression suite captures each reviewed edge case with no-write assertions where relevant.
- Scaffold propagation and policy mirror checks remain covered and green.

## Merge Recommendation
Approve

## Top Risks
- No merge-blocking risks found.
- The superseded snapshot behavior remains intentionally manual/by design, as documented in the feedback response.

## Suggested Additional Tests
- None blocking.

PR #133, opened 2026-07-14
Fix AGENTS.md discovery on BSD find in non-git repos

ssheld:

## Summary

`check-memory-budget.sh`'s non-git AGENTS.md discovery fallback used GNU-only `find -printf '%P\n'`, which BSD `find` on macOS rejects (`-printf: unknown primary or operator`). The failure was doubly swallowed — stderr discarded, and the pipeline runs inside a process substitution whose exit status is never checked — so the Codex AGENTS bucket silently reported `MISSING (no AGENTS.md found)` and an over-budget `AGENTS.md` never tripped `--strict`. This is exactly the path the tracked pre-commit hook exercises: it materializes the index via `git checkout-index` into a temp directory (no `.git`), then runs `check-memory-budget.sh --repo <staged_root> --strict`. On macOS the staged memory-budget warning could therefore never fire for an AGENTS.md overage.

Found by a Cursor Grok 4.5 review of a downstream scaffold sync in a private consumer repo; reproduction confirmed independently before fixing.

## Files / areas changed

- `scaffold/root/scripts/check-memory-budget.sh`: replace the GNU-only `-printf '%P\n'` with a portable `find . | sed 's|^\./||' | sort -u` pipe. The `sed` strip is load-bearing: a bare `-print` emits `./AGENTS.md`, which still measures bytes correctly but breaks exceptions-file matching (keyed by clean repo-relative paths) and display parity with the git branch.
- `scripts/test-check-memory-budget.sh`: the existing non-git discovery assertion (`sub/AGENTS.md`) failed on stock macOS before this fix and passes after. Added: root `AGENTS.md` must be discovered without a `./` prefix (guards against a naive `-print`-only fix), and an over-budget AGENTS.md in a non-git directory must fail `--strict` with `over file budget` (pins the hook's exact scenario).

## Verification

Run locally (macOS, BSD userland, Homebrew bash 5.3):
- `bash scripts/test-check-memory-budget.sh` — pass (failed on `main` on this machine at the non-git discovery assertion before the fix).
- `bash scripts/check-style.sh` — pass (ShellCheck + shfmt).
- Manual repro: non-git dir with a 45,000-byte `agent-vault/AGENTS.md` (budget 40,000) — before: `MISSING` + `Within budget` + exit 0; after: `OVER ... over file budget (40000 bytes)` + strict exit 1.

## Risks / rollback

- Risk: low — one-line behavioral change confined to the non-git fallback branch; the git-repo branch (`git ls-files`, incl. `sort -zu`) is untouched and verified working on macOS.
- Residual gap: CI (`scaffold-regression-checks.yml`) runs on `ubuntu-latest` with GNU find, which is why this never surfaced there. Worth considering a `macos-latest` job for the shell suites as a follow-up; not included to keep this PR scoped.
- Rollback: revert the single commit.

## Docs Consistency

- `docs/memory-budgets.md` describes the fallback behaviorally ("a plain filesystem walk") — still accurate; no update needed.
- `README.md`: no impact (no CLI or usage change).

### DIFF
--- a/scaffold/root/scripts/check-memory-budget.sh
+++ b/scaffold/root/scripts/check-memory-budget.sh
@@ -306,7 +306,9 @@ discover_agents_files() {
     git -C "$repo" -c core.quotePath=false ls-files -z --cached --others --exclude-standard \
       -- 'AGENTS.md' '**/AGENTS.md' 2>/dev/null | sort -zu | tr '\0' '\n'
   else
-    (cd "$repo" && find . -name AGENTS.md -not -path '*/.git/*' -printf '%P\n' 2>/dev/null | sort -u)
+    # BSD find has no -printf; strip the leading "./" so paths stay
+    # repo-relative for exception matching and display parity with git ls-files.
+    (cd "$repo" && find . -name AGENTS.md -not -path '*/.git/*' 2>/dev/null | sed 's|^\./||' | sort -u)
   fi
 }
 
--- a/scripts/test-check-memory-budget.sh
+++ b/scripts/test-check-memory-budget.sh
@@ -259,6 +259,17 @@ printf 'root\n' >"$nongit/AGENTS.md"
 printf 'nested\n' >"$nongit/sub/AGENTS.md"
 mapfile -t nongit_agents < <(tsv_bucket "$nongit" agents)
 assert_in_set "sub/AGENTS.md" "${nongit_agents[@]}"
+# The root file must appear WITHOUT a "./" prefix: exception matching and the
+# git-branch output are keyed by clean repo-relative paths (regression: the
+# GNU-only "find -printf '%P\n'" fallback returned nothing on BSD userland).
+assert_in_set "AGENTS.md" "${nongit_agents[@]}"
+assert_not_in_set "./AGENTS.md" "${nongit_agents[@]}"
+
+# The pre-commit hook's staged-checkout path is a non-git directory: an
+# over-budget AGENTS.md there must still trip --strict.
+head -c 45000 /dev/zero | tr '\0' 'a' >"$nongit/sub/AGENTS.md"
+expect_result 1 "over file budget" --repo "$nongit" --strict
+printf 'nested\n' >"$nongit/sub/AGENTS.md"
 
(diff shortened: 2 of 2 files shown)


ssheld (commented):
> 🤖 **Review by Cursor Grok 4.5** · via Cursor CLI
Formal review state unavailable: GitHub disallows approving your own pull request.

## Strengths
- Portable `find | sed 's|^\./||' | sort -u` fallback correctly replaces GNU-only `-printf '%P\n'` and preserves clean repo-relative paths for exception matching and display parity with the git discovery branch.
- Regression coverage matches the failure mode: root `AGENTS.md` without a `./` prefix, rejection of `./AGENTS.md`, and `--strict` failure on an over-budget AGENTS file in a non-git directory (the `checkout-index` staged-temp path).
- Verified locally on macOS: suite passes, `check-style.sh` passes, and a non-git fixture with a 45KB `agent-vault/AGENTS.md` reports `OVER` and exits 1 under `--strict`.

## Merge Recommendation
Approve

## Top Risks
- CI remains on `ubuntu-latest` (GNU find), so similar BSD/macOS shell assumptions could regress unnoticed; the PR already calls this out as a scoped follow-up, which is the right split.

## Suggested Additional Tests
- None required for merge. A later `macos-latest` job for the shell suites would harden the residual CI gap noted in the PR body.

ssheld (commented):

## Strengths

- Replacing GNU-only `find -printf` with the portable `find | sed | sort` path restores non-git discovery on BSD userland while preserving clean repo-relative names for exception matching.
- The regressions cover root and nested `AGENTS.md` discovery, reject an accidental `./` prefix, and prove that a staged-checkout overage fails under `--strict`.
- The standalone checker, pre-commit hook regression, shell style checks, and both GitHub checks pass at `cbc15f7`.

## Merge Recommendation

Approve

## Top Risks

- BSD-specific behavior is still validated locally rather than by a macOS CI job, so a future portability regression would depend on review or local testing to catch it.

## Suggested Additional Tests

- No additional test is required for this patch. A `macos-latest` shell-test job would be a useful separate CI-hardening follow-up.

PR #134, opened 2026-07-14
Restrict rollover entry boundaries to canonical entry headings

ssheld:

## Summary

`compact-context-log.sh` and `check-context-log-rollover.sh` accepted **any** Markdown heading (any depth) whose text merely starts with a `YYYY-MM-DD` date as an entry boundary. A nested sub-heading like `#### 2026-05-29 follow-up` inside a real entry was therefore counted as another entry, with two failure modes (both reproduced before fixing):

1. **Mid-entry split**: when the nested heading lands at the keep boundary, the compactor tears its parent entry apart — the tail goes into the archive as an orphaned `####` fragment, the live copy silently loses content, and the manifest records the fragment as `newest_archived` with an inflated `archived` count. Because the checker used the same loose matcher, both the compactor's self-validation and a subsequent standalone check passed the corrupted result.
2. **Inflated trigger**: nested dated headings count toward `total_entries`, so a log with exactly `--keep` real entries rolls over when it should be a no-op — archiving *only* the orphaned fragment.

Both scripts now match the canonical `### YYYY-MM-DD HH:MM local - <agent> - <topic>` shape — the same form the tracked pre-commit hook already enforces on live entries (the hook's own matcher was already strict and is unchanged).

Found by a Codex review of a downstream scaffold sync in a private consumer repo; reproduction confirmed independently, including the no-op corruption the review did not mention.

## Files / areas changed

- `scaffold/root/scripts/compact-context-log.sh`: strict matcher in all three heading scanners — `entry_heading_lines` (split boundaries + entry count), `first_entry_line` (archive header slicing), and `select_boundaries` (manifest `newest_archived`/`oldest_archived`). The `entry_ts` 00:00 fallback became dead code under the guaranteed format and was removed (`ts = substr(line, 1, 16)`).
- `scaffold/root/scripts/check-context-log-rollover.sh`: strict matcher in `verify_archive_boundaries`, plus help text and comments — the old help explicitly documented "Entry heading styles inside `## Entries` are not constrained", which the fix supersedes for boundary verification.
- `scripts/test-compact-context-log.sh`: `count_entries` helper tightened to the same shape; new regressions — a nested dated sub-heading must stay with its kept entry (whole-entry split, correct counts, archive must not start with `####`, checker passes), and the exact-keep no-op must write nothing.
- `scripts/test-check-context-log-rollover.sh`: new regressions — a nested dated sub-heading inside an archived entry does not disturb boundary verification (positive), and a fragment-topped archive whose manifest cites the fragment as `newest_archived` fails with `newest_archived heading not found` (fail-closed).

**Review follow-up (`4d3263b`)** — closes the fail-open hole both reviews converged on:
- `scaffold/root/scripts/check-context-log-rollover.sh`: archives are now checked for dated headings the strict boundary scan cannot see — any dated heading *above* the first canonical entry is flagged as a torn-entry fragment, and any depth-1..3 noncanonical dated heading (legacy `##` style, em-dash, missing `HH:MM local -`) is flagged for normalization. Dated sub-headings of depth 4+ inside an entry remain legal body text. Help text updated.
- `scaffold/root/scripts/compact-context-log.sh`: an existing archive is validated with the checker **before any writes**; legacy/torn archives abort with a normalization pointer instead of being silently treated as header prose and reordered above the newer batch. Deliberately no override flag — an override would recreate the silent-reorder hole; normalizing the archive is the fix.
- Tests: the nested-fixture orphan assertion was hollow (`head -c 4` could never see a fragment below the `# Context Log Archive` header) and is replaced by asserting the archive's first dated heading is the canonical boundary entry; new regressions pin the legacy-only archive abort (zero writes, archive and log unchanged, no manifest), the mixed noncanonical-below-canonical abort, and checker failures for noncanonical and header-fragment archives (no manifest required).

## Design note

Full canonical-shape matching was chosen over the softer depth-3-only alternative because it aligns these scripts with the hook's enforced format and is fail-closed: a malformed heading in an archive becomes invisible to boundary selection, producing a manifest mismatch → checker failure, rather than silent corruption. Legacy em-dash (`local —`) headings live only under `## Historical Indexed Entries` wrappers outside `## Entries`, which these scanners never enter.

## Verification

Run locally (macOS, Homebrew bash 5.3):
- `bash scripts/test-compact-context-log.sh` — pass (130 assertions, including the new regressions).
- `bash scripts/test-check-context-log-rollover.sh` — pass.
- `bash scripts/test-session-metadata-hook.sh`, `bash scripts/test-memory-budget-hook.sh` — pass (hook suites exercising these scripts).
- `bash scripts/check-style.sh` — pass (ShellCheck + shfmt).
- Manual repro of both failure modes on `main`, re-run after the fix: whole entries move, counts correct, fragment archives rejected.

## Risks / rollback

- Risk: an existing hand-written or legacy archive containing noncanonical dated entry headings now fails the checker and blocks compaction with an actionable message, instead of being silently reordered (pre-review behavior) or loosely counted (pre-PR behavior). This is intended fail-closed behavior; such archives must be normalized once. Compactor-produced archives are canonical by construction.
- Rollback: revert the branch (three commits).

## Docs Consistency

- `README.md` / `docs/`: no references to the loose matcher semantics found; the checker's own `--help` text was updated in this PR.

### DIFF
--- a/scaffold/root/scripts/check-context-log-rollover.sh
+++ b/scaffold/root/scripts/check-context-log-rollover.sh
@@ -26,6 +26,11 @@ Checks (live context log; headings matched outside fenced code blocks):
   - if the snapshot declares a latest-handoff pointer, it has an inline value
 Checks (when --archive is given):
   - every archived "## Current Snapshot" is labeled superseded
+  - no dated heading sits above the archive's first canonical entry heading
+    (a torn entry fragment), and every dated heading of depth 1-3 is a
+    canonical entry heading (noncanonical legacy entries would be invisible
+    to boundary validation); dated sub-headings of depth 4+ inside an entry
+    remain legal body text
 Checks (when --manifest is given -- Layer-2 rollover assertions):
   - the live Current Snapshot's "Context-log rollover" pointer references the
     newest manifest record's id and repeats its boundary text verbatim
@@ -205,6 +210,45 @@ inspect_archive_superseded() {
   ' "$1"
 }
 
+# Prints "<line>\t<kind>\t<heading>" for archive dated headings (outside fences)
+# that the strict boundary scan cannot treat as entries:
+#   fragment     -- any dated heading before the first canonical entry heading;
+#                   only header prose belongs there, so this is the signature of
+#                   an entry torn apart by a bad split (or a misplaced paste)
+#   noncanonical -- a depth-1..3 heading whose text starts with a YYYY-MM-DD
+#                   date but is not canonical (legacy or hand-written entry the
+#                   boundary scan would silently skip)
+# Dated sub-headings of depth 4+ inside an entry are legal body text.
+inspect_archive_dated_headings() {
+  awk '
+    function strip(s) {
+      sub(/\r$/, "", s)
+      sub(/^[[:space:]]+/, "", s)
+      sub(/[[:space:]]+$/, "", s)
+      return s
+    }
+    /^(```|~~~)/ { in_fence = !in_fence; next }
+    {
+      if (in_fence) next
+      line = strip($0)
+      if (line !~ /^#+[[:space:]]/) next
+      text = line
+      sub(/^#+[[:space:]]+/, "", text)
+      if (text !~ /^[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]/) next
+      if (line ~ /^### [0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] [0-9][0-9]:[0-9][0-9] local - /) {
+        seen_entry = 1
+        next
+      }
+      if (!seen_entry) {
+        printf "%d\tfragment\t%s\n", NR, line
+        next
+      }
+      match(line, /^#+/)
+      if (RLENGTH <= 3) printf "%d\tnoncanonical\t%s\n", NR, line
+    }
+  ' "$1"
+}
+
 # --- Layer-2 (rollover manifest) helpers ---------------------------------
 # These run only when --manifest is given. They verify the live pointer's claim
 # against the manifest, and the manifest's claims against archive reality, so a
@@ -362,6 +406,18 @@ if [[ -n "$archive_file" ]]; then
     [[ -n "$row" ]] || continue
     findings+=("archived snapshot is not labeled superseded (archive line ${row%%$'\t'*}) -- archived snapshots must be marked superseded so they cannot read as active")
   done < <(inspect_archive_superseded "$archive_file")
+
+  while IFS=$'\t' read -r row_line row_kind row_heading; do
+    [[ -n "$row_line" ]] || continue
+    case "$row_kind" in
+      fragment)
+        findings+=("archive has a dated heading above its first entry (archive line $row_line): \"$row_heading\" -- looks like a torn entry fragment; only header prose belongs above the first canonical \"### YYYY-MM-DD HH:MM local - ...\" heading")
+        ;;
+      noncanonical)
+        findings+=("archive has a noncanonical dated entry heading (archive line $row_line): \"$row_heading\" -- normalize it to \"### YYYY-MM-DD HH:MM local - <agent> - <topic>\" so boundary validation can see it")
+        ;;
+    esac
+  done < <(inspect_archive_dated_headings "$archive_file")
 fi
 
 if [[ -n "$manifest_file" ]]; then
--- a/scaffold/root/scripts/check-context-log-rollover.sh
+++ b/scaffold/root/scripts/check-context-log-rollover.sh
@@ -36,7 +36,9 @@ Checks (when --manifest is given -- Layer-2 rollover assertions):
 
 Section headings are matched exactly (a distinct heading such as
 "## Current Snapshot Format Notes" is not a duplicate), and CRLF line endings
-are tolerated. Entry heading styles inside "## Entries" are not constrained.
+are tolerated. Live entry-heading style is enforced by the pre-commit hook, not
+here; archive boundary verification counts only canonical
+"### YYYY-MM-DD HH:MM local - <agent> - <topic>" entry headings.
 
 The rollover manifest (parsed source of truth) holds one record per rollover,
 newest first. All fields are required; the *_archived headings are the entry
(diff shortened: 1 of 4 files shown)


ssheld (commented):
> 🤖 **Review by Cursor Grok 4.5** · via Cursor CLI
Formal review state unavailable: GitHub disallows requesting changes on your own pull request.

1. `scripts/test-compact-context-log.sh:682`

 **Recommended** · High confidence — The orphan-fragment assertion does not detect the failure mode it describes.
 `[[ "$(head -c 4 "$d/archive/context-log-2026.md")" != "####" ]]` checks the first four bytes of the archive file. Compactor-built archives always begin with `# Context Log Archive`, so those bytes are `# Co` even when the archived *batch* starts with an orphaned `#### <date>` fragment under that header. The real mid-entry split regression is still covered by the kept-count message, the live-file `#### 2026-05-29 follow-up` retention check, and the `newest_archived` expectation — but this particular pin is hollow. Prefer asserting that the first entry heading in the archive is canonical (for example, the first non-fence `^### ` / `^#### ` line must match the strict entry shape), or that the archive body does not contain a leading `^#### [0-9]{4}-` entry fragment after the header.

## Strengths
- Correct root-cause fix: aligning the compactor and checker with the hook's canonical `### YYYY-MM-DD HH:MM local - ...` shape closes both the mid-entry split and the inflated `--keep` trigger, including the fail-closed path where a fragment-topped archive no longer self-validates.
- Fence-awareness is exercised in the nested fixture, so a fenced canonical heading cannot become a split boundary.
- Local verification is solid: compact (121 assertions), rollover checker, session-metadata hook, memory-budget hook, and `check-style.sh` all passed on this branch.

## Merge Recommendation
Approve with Changes

## Top Risks
- A future edit could weaken the effective nested-heading assertions while leaving the hollow `head -c 4` check in place, giving false confidence that the orphan-fragment case is still pinned.

## Suggested Additional Tests
- Replace the `head -c 4` check with an assertion over the first dated heading line in the archive (canonical `### ... local - ` required; `#### <date>` must fail).
- Optional: a dedicated fixture where the archive header is omitted/`first_entry_line == 1`, so a fragment at byte 0 would also be covered if that path remains supported.

ssheld (commented):

1. `scaffold/root/scripts/compact-context-log.sh:225`
   `scaffold/root/scripts/compact-context-log.sh:384`

   **Recommended** · High confidence — Fail closed when an existing archive contains noncanonical dated entry headings.
   `first_entry_line` now returns no match unless the existing archive already uses the full canonical shape. The caller interprets that as a header-only archive, places all existing content in `arch_header`, and appends the newly archived canonical batch beneath it. A hand-written archive with headings such as `## 2026-05-29 ...` or a missing `local -` segment is therefore silently kept above newer entries, ignored by `select_boundaries`, and ignored by the checker, so self-validation passes instead of producing the intended normalization failure. Detect date-prefixed but noncanonical archive headings and abort before writes, or otherwise ensure they cannot be mistaken for header prose.

## Strengths

- Canonical matching fixes both the mid-entry split and inflated rollover-trigger defects.
- The new regressions cover nested dated body headings, fenced canonical-looking examples, and the exact-`--keep` no-op.
- Focused compactor, checker, hook, and style suites pass at the PR head, and CI is green.

## Merge Recommendation

Approve with Changes

## Top Risks

- Existing hand-written or legacy archives can be silently reordered and excluded from manifest boundary validation.

## Suggested Additional Tests

- Add an existing archive containing only noncanonical dated headings and assert an actionable failure with zero writes.
- Add a mixed archive with noncanonical headings before canonical entries and assert the same fail-closed behavior.

ssheld:
> 🤖 **Feedback response by Claude Fable 5** · via Claude Code

All feedback surfaces retrieved with pagination (top-level reviews, inline review comments, conversation comments): two formal reviews (Cursor Grok 4.5, Codex), no inline or conversation comments. Both findings were verified against the code before changing anything — both confirmed. Addressed in `4d3263b`; all four affected suites pass locally (compactor now 130 assertions, rollover checker, session-metadata hook, memory-budget hook), `check-style.sh` clean, CI green on the PR head.

| 1 | Cursor: `scripts/test-compact-context-log.sh:682` — the `head -c 4` orphan-fragment assertion is hollow (compactor archives always begin with `# Context Log Archive`, so the first four bytes can never be `####`) | Resolved | Confirmed: new archives are created with that exact header, so the check was unfalsifiable. Replaced per the review's suggestion with a fence-aware scan asserting the archive's **first dated heading** equals the canonical boundary entry (`### 2026-05-28 10:00 local - codex - older work`) — a `#### <date>` fragment or a reordered batch now surfaces there directly. `4d3263b`, `scripts/test-compact-context-log.sh`. The "optional" headerless/byte-0 case is covered by the new checker rule from finding 2: any dated heading above the archive's first canonical entry now fails validation regardless of manifest involvement (pinned in `scripts/test-check-context-log-rollover.sh`). |
| 2 | Codex: `scaffold/root/scripts/compact-context-log.sh:225,384` — strict `first_entry_line` turns a noncanonical existing archive into "header prose": silently reordered above the newer batch and invisible to `select_boundaries` and the checker | Resolved | Confirmed against the archive-build path (empty `first_entry_line` → whole file into `arch_header`, batch appended below). Fixed fail-closed at two layers in `4d3263b`: (a) `check-context-log-rollover.sh` now flags archive dated headings the boundary scan cannot see — any dated heading above the first canonical entry ("torn entry fragment") and any depth-1..3 noncanonical dated heading (legacy `##` style, em-dash, missing `HH:MM local -`), while depth-4+ dated sub-headings inside entries stay legal body text (existing positive regression still passes); (b) `compact-context-log.sh` validates an existing archive with the checker **before any writes** and aborts with a normalization pointer — and because self-validation reuses the checker, the built archive is guarded by the same rules. Deliberately no override flag: an override would recreate the silent-reorder hole, and the stale-archive-metadata precedent's escape hatch doesn't fit here since normalization is cheap and one-time. Both suggested tests added: legacy-only archive → abort, exit 1, zero writes (log, archive, and absent manifest all verified unchanged); mixed noncanonical-below-canonical archive → same abort. Checker-side failures pinned without any manifest (`noncanonical dated entry heading`, `above its first entry`). |

## Remaining Items
- None. PR body updated to reflect the follow-up commit (Files/areas changed, Verification counts, and the Risks section now describes the abort-with-pointer behavior for legacy archives — they must be normalized once before compaction).

ssheld (commented):

## Feedback Coverage

- **Resolved — hollow orphan-fragment assertion.** Commit `4d3263b` replaces the byte-zero check with a fence-aware assertion over the archive’s first dated heading in `scripts/test-compact-context-log.sh`.
- **Resolved — noncanonical archives could be treated as header prose.** Commit `4d3263b` adds archive-side detection in `scaffold/root/scripts/check-context-log-rollover.sh`, preflight validation before writes in `scaffold/root/scripts/compact-context-log.sh`, and legacy-only plus mixed-archive zero-write regressions.
- Reviewed both prior formal reviews, the itemized feedback-response comment, and inline thread state. No inline threads or unaddressed actionable comments remain.

## Strengths

- The original nested-heading corruption and inflated-trigger paths remain covered.
- Existing legacy or torn archives now fail closed before any destination mutation.
- The compactor reports 130 passing assertions; checker and hook suites pass; shell style and both GitHub checks are green at `4d3263b`.

## Merge Recommendation

Approve

## Top Risks

- The intentional compatibility cost is explicit: noncanonical legacy archives must be normalized once before compaction can proceed.

## Suggested Additional Tests

- None required for this change; the requested positive, negative, and zero-write regression paths are covered.

PR #136, opened 2026-07-16
ci: run scaffold regression suites on macOS as well as Ubuntu

ssheld:

## Summary

The scaffold regression suites ran only on `ubuntu-latest` (GNU userland). That platform gap let a BSD-incompatible `find -printf` in `check-memory-budget.sh` ship even though an existing test asserted exactly the behavior it broke — the suite failed on any stock macOS machine, but CI never executed it there. Fixed in #133; both #133 reviews endorsed a `macos-latest` job as the follow-up hardening. This PR adds it.

## Files / areas changed

- `.github/workflows/scaffold-regression-checks.yml`: the job becomes a two-leg matrix (`ubuntu-latest`, `macos-latest`) with `fail-fast: false` so one platform's failure cannot cancel the other's evidence — cross-platform divergence is precisely the signal this exists to capture. A macOS-only first step ensures bash 4+ (Apple ships 3.2 at `/bin/bash`; the suites use `mapfile`/`declare -A`), installing Homebrew bash only if the runner image lacks it. All 18 verify steps are unchanged and run identically on both legs.
- `README.md`: the Scaffold Regression Checks section now documents the two-platform CI run and the local bash 4+ requirement on macOS.

No check-name pinning concerns: the active "Protect main branch" ruleset requires PRs but names no required status checks, so the matrix rename of the check contexts is safe.

## Verification

- All 18 suites pass locally on macOS (BSD userland, Homebrew bash 5.3) — the exact environment the new leg represents; the memory-budget suite among them failed on `main` before #133, demonstrating the leg would have caught that class of bug.
- Workflow YAML parses cleanly; the matrix leg proves itself on this PR's own checks (the paths filter includes the workflow file).
- `bash scripts/check-style.sh` — pass (no shell files changed; run as regression hygiene).

## Risks / rollback

- Risk: low — additive CI leg; Ubuntu leg and step list unchanged. macOS runners are slower, so wall-clock for the workflow increases; public-repo hosted runners keep this free.
- Rollback: revert the single commit (drops the matrix back to Ubuntu-only).

## Docs Consistency

- `README.md` updated in this PR; no other docs describe CI platforms.

### DIFF
--- a/.github/workflows/scaffold-regression-checks.yml
+++ b/.github/workflows/scaffold-regression-checks.yml
@@ -69,10 +69,25 @@ on:
 
 jobs:
   scaffold-regression-checks:
-    runs-on: ubuntu-latest
+    strategy:
+      # Run both platforms to completion: BSD-vs-GNU userland differences are
+      # exactly what the macOS job exists to catch (a GNU-only `find -printf`
+      # once made AGENTS.md discovery silently return nothing on macOS), so an
+      # Ubuntu failure must not cancel the macOS evidence or vice versa.
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Ensure bash 4+ on macOS (Apple ships bash 3.2; suites use mapfile/declare -A)
+        if: runner.os == 'macOS'
+        run: |
+          if ! bash -c '[[ ${BASH_VERSINFO[0]} -ge 4 ]]'; then
+            brew install bash
+          fi
+          bash --version | head -1
       - name: Verify managed .gitignore handling
         run: bash scripts/test-gitignore-management.sh
       - name: Verify coding standards sync behavior
--- a/.github/workflows/scaffold-regression-checks.yml
+++ b/.github/workflows/scaffold-regression-checks.yml
@@ -83,11 +83,26 @@ jobs:
         uses: actions/checkout@v4
       - name: Ensure bash 4+ on macOS (Apple ships bash 3.2; suites use mapfile/declare -A)
         if: runner.os == 'macOS'
+        # Assert against the exact Homebrew binary, not bare "bash": this
+        # step's shell hashes "bash" to /bin/bash on first use, so a bare
+        # post-install "bash --version" reports 3.2 even after a successful
+        # install and cannot prove the postcondition.
         run: |
-          if ! bash -c '[[ ${BASH_VERSINFO[0]} -ge 4 ]]'; then
+          brew_bin="$(brew --prefix)/bin"
+          if [ ! -x "$brew_bin/bash" ]; then
             brew install bash
           fi
-          bash --version | head -1
+          echo "$brew_bin" >> "$GITHUB_PATH"
+          "$brew_bin/bash" -c '[[ ${BASH_VERSINFO[0]} -ge 4 ]]'
+          "$brew_bin/bash" --version | head -1
+      - name: Verify the suite interpreter is bash 4+ (macOS)
+        if: runner.os == 'macOS'
+        # Fresh step, fresh shell: bare "bash" here resolves exactly as it
+        # will in every suite step below, so this asserts what the suites
+        # actually run with (not just that a modern bash exists somewhere).
+        run: |
+          command -v bash
+          bash -c 'echo "suite bash: $BASH_VERSION"; [[ ${BASH_VERSINFO[0]} -ge 4 ]]'
       - name: Verify managed .gitignore handling
         run: bash scripts/test-gitignore-management.sh
       - name: Verify coding standards sync behavior
--- a/README.md
+++ b/README.md
@@ -144,7 +144,11 @@ Run the scaffold regression scripts locally when changing bootstrap, sync, or tr
 - `bash scripts/test-remove-worktree.sh`
 - `bash scripts/test-worktree-helper-sync.sh`
 
-CI also runs these checks via `.github/workflows/scaffold-regression-checks.yml`.
+CI runs these checks via `.github/workflows/scaffold-regression-checks.yml` on
+both `ubuntu-latest` and `macos-latest`, so GNU-vs-BSD userland assumptions in
+the shell scripts fail in CI instead of surfacing on contributor machines.
+Locally the suites need bash 4+ (macOS ships bash 3.2 at `/bin/bash`; use
+Homebrew bash).
 
 ## Updating Existing Repos
 `update-project.sh` updates these managed scaffold files:


ssheld (commented):
Formal review state unavailable: GitHub does not allow authors to approve or request changes on their own pull requests.

1. `.github/workflows/scaffold-regression-checks.yml:90`

   **Recommended** · High confidence — Verify the post-install Bash interpreter.
   The macOS Actions log for this PR ends the new setup step with `GNU bash, version 3.2.57` even after Homebrew installs Bash 5.3. The initial probe causes the current shell to cache `/bin/bash`, and `bash --version | head -1` only prints output, so the “Ensure bash 4+” step can succeed without proving its postcondition. Later steps happened to re-resolve Homebrew Bash (the suites using `mapfile` passed), but that depends on step-boundary command resolution and leaves misleading setup evidence. Reset command lookup with `hash -r` or invoke `$(brew --prefix)/bin/bash` explicitly, then make the version test an assertion; persist the selected path through `$GITHUB_PATH` if later steps are intended to use bare `bash`.

## Strengths

- The two-leg matrix preserves the existing Ubuntu coverage and `fail-fast: false` keeps both platform results available.
- The README accurately explains the GNU/BSD coverage goal and the local Bash requirement.
- Both matrix legs pass, and the macOS leg exercises all 18 regression suites.

## Merge Recommendation

Approve with Changes

## Top Risks

- A future runner-image or `PATH` change could leave the macOS job using Apple Bash while the setup step still reports success.

## Suggested Additional Tests

- Re-run the matrix after making the setup step assert and log Bash 4+ from the exact interpreter later suite commands will resolve.

ssheld (commented):
> 🤖 **Review by Cursor Grok 4.5** · via Cursor

1. `.github/workflows/scaffold-regression-checks.yml:84`

 **Recommended** · High confidence — macOS bash ensure step does not prove bash 4+ is what later steps use.

 The new step correctly detects Apple's bash 3.2 and runs `brew install bash`, but its post-install check is not a real gate: on this PR's macOS job it still printed `GNU bash, version 3.2.57(1)-release` after the install completed. Later suite steps happened to switch to `shell: /opt/homebrew/bin/bash`, so the matrix passed, but that depends on Actions re-resolving the default shell between steps rather than on an explicit PATH/assertion in this workflow.

 The step should (a) put Homebrew's bin on `GITHUB_PATH` so subsequent `bash scripts/...` invocations reliably prefer the installed binary, and (b) assert bash ≥ 4 against that binary (e.g. `"$(brew --prefix)/bin/bash"`) so a failed install cannot greenwash. Suggested shape:

 ```yaml
 - name: Ensure bash 4+ on macOS (Apple ships bash 3.2; suites use mapfile/declare -A)
   if: runner.os == 'macOS'
   run: |
     if ! bash -c '[[ ${BASH_VERSINFO[0]} -ge 4 ]]'; then
       brew install bash
     fi
     echo "$(brew --prefix)/bin" >> "$GITHUB_PATH"
     "$(brew --prefix)/bin/bash" -c '[[ ${BASH_VERSINFO[0]} -ge 4 ]]'
     "$(brew --prefix)/bin/bash" --version | head -1
 ```

## Strengths
- Matrix + `fail-fast: false` is the right shape for catching BSD-vs-GNU divergence without one platform cancelling the other.
- Scope stays tight: same 18 suites on both legs, with README updated to document the bash 4+ local requirement.
- Confirmed the active "Protect main branch" ruleset requires PRs but pins no required status checks, so the matrix check-context rename is safe.

## Merge Recommendation
Approve with Changes

## Top Risks
- Ensure-step false confidence: install can succeed while the logged `bash --version` still shows 3.2, masking PATH/hashing issues if Actions ever stops re-picking Homebrew bash for later steps.
- macOS leg is ~4–5× slower than Ubuntu (~3m vs ~40s on this PR); acceptable for public runners, but worth watching if the suite list grows.

## Suggested Additional Tests
- After the ensure step, assert `command -v bash` resolves under `$(brew --prefix)/bin` (or print `BASH_VERSINFO` from a follow-up step that uses the same `bash scripts/...` invocation pattern as the suites).

ssheld:
> 🤖 **Feedback response by Claude Fable 5** · via Claude Code

All feedback surfaces retrieved with pagination (top-level reviews, inline review comments, conversation comments): two formal reviews (Codex, Cursor Grok 4.5), no inline or conversation comments. Both reviews report the same underlying defect at `.github/workflows/scaffold-regression-checks.yml:84–90`, so the duplicate findings are grouped into one row and fixed once.

| 1 | Codex + Cursor (grouped): the "Ensure bash 4+" step cannot prove its postcondition — the initial bare `bash -c` probe hashes `/bin/bash` in the step's shell, so the post-install `bash --version` printed `3.2.57` even after Homebrew installed 5.3, and later suites got a modern bash only via incidental step-boundary PATH re-resolution | Resolved | Confirmed against the prior run's log before fixing (ensure step ended with `GNU bash, version 3.2.57(1)-release`). Fixed in `e9823ce`: the step now selects the interpreter deterministically (`$(brew --prefix)/bin/bash`, installing only if that binary is absent — no bare-`bash` probe left to hash-poison the shell), asserts `BASH_VERSINFO[0] >= 4` against that exact binary so a failed install fails the step instead of greenwashing, and persists the prefix via `$GITHUB_PATH`. Both suggested verifications implemented: (a) Cursor's follow-up assertion is a new fresh-shell step that checks `command -v bash` and asserts the version through the same bare-`bash` invocation pattern every suite step uses; (b) Codex's "re-run the matrix" — done, all legs green. Post-fix macOS log evidence: ensure step prints `GNU bash, version 5.3.15(1)-release`; verify step prints `/opt/homebrew/bin/bash` and `suite bash: 5.3.15(1)-release`; every subsequent suite step runs with `shell: /opt/homebrew/bin/bash -e {0}`. |

## Remaining Items
- None. Cursor's runtime observation (macOS leg ~4–5× slower than Ubuntu) is noted; acceptable on public hosted runners and worth revisiting only if the suite list grows materially.

ssheld (commented):
Formal review state unavailable: GitHub does not allow authors to approve their own pull requests.

The two prior reviews raised the same Bash-interpreter issue. Commit `e9823ce` fully addresses it: the workflow selects and asserts the exact Homebrew Bash binary, persists its directory through `$GITHUB_PATH`, and verifies in a fresh step that bare `bash` resolves to Bash 4+ before any suite runs. The latest macOS log confirms Bash 5.3.15 at `/opt/homebrew/bin/bash` for the verification step and every one of the 18 regression suites.

## Strengths

- The feedback fix proves both installation and downstream interpreter selection, eliminating the previous false-positive setup check.
- The Ubuntu leg remains unchanged in behavior while `fail-fast: false` preserves complete cross-platform evidence.
- Documentation remains aligned with the CI behavior and local macOS requirement.

## Merge Recommendation

Approve

## Top Risks

- The macOS leg depends on Homebrew availability and is slower than Ubuntu, but installation or interpreter failures now fail observably instead of silently using Bash 3.2.

## Suggested Additional Tests

- None. The latest Style Check and both regression-matrix legs pass on `e9823ce`; I also reran shell style, workflow YAML parsing, `git diff --check`, and all 18 regression suites locally under Homebrew Bash 5.3.

